PR #1, opened 2022-12-07
Add skeleton for the ZBC Solidity library

dartdart26:
Separate into two sub-libraries:
 * Ciphertext - for managing ciphertexts
 * FHEOperations - for homomorphic operations on ciphertexts

The exposed library functions just dispatch the calls to the respective precompiled contracts.

An attempt is made to make it easy for developers to use the Zama Blockchain through simple and clear interfaces.

### DIFF
--- a/contracts/FHEOperations.sol
+++ b/contracts/FHEOperations.sol
@@ -2,8 +2,9 @@
 
 pragma solidity >=0.7.0 <0.9.0;
 
-// A library of functions for homomorphic operations on ciphertexts.
+import "./Common.sol";
 
+// A library of functions for homomorphic operations on ciphertexts.
 // NOTE: Currently, all handles refer to ciphertexts of the same type, i.e. an 8-bit integer. This
 // is likely to change in the future.
 library FHEOperations {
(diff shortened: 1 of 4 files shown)


contracts/FHEOperations.sol
```diff
@@ -0,0 +1,31 @@
...
+    function add(uint256 handleA, uint256 handleB) internal view returns (uint256 resultHandle) {
+        bytes32[2] memory input;
+        input[0] = bytes32(handleA);
+        input[1] = bytes32(handleB);
+
+        bytes32[1] memory output;
+
+        // Call the add precompile.
+        assembly {
+            if iszero(staticcall(gas(), 65, input, 64, output, 32)) {
```
mortendahl: what are the hard coded numbers? could we pull them into a constant?

mortendahl (commented):
👍

contracts/FHEOperations.sol
dartdart26: These are the addresses of the precompiled contracts. It's a good idea - will put them as constants in one place. I just forgot to do it...

PR #10, opened 2023-01-29
feat() add field and rename some method in example

immortal-tofu:


tremblaythibaultl (approved):
lgtm 👍
Demo test when through no problem : http://13.36.253.113:4000/address/0x28D8E129A1a46c775BA9F67A2d0a7FbC6c22c47e.

PR #13, opened 2023-02-02
Add a blind auction example

dartdart26:
The `BlindAuction` example contract supports a blind auction with the following properties:
 * bids remain hidden
 * winning bid remains hidden
 * winner address becomes public at the end

In order to support hidden bids, we link the auction to the `EncryptedERC20` contract, such that bids are in this token. We utilize the allowances mechanism to approve use of user tokens on their behalf by the auction contract.

The flow is:
1. Auction owner creates the auction contract.
2. Users approve use of tokens in the ERC20 contract via `approve()`.
3. Users call `bid()` to bid an amount less than or equal to the approved token amount.
4. Auction owner stops the auction.
5. Users can call `doIHaveHighestBid() to know if they have the highest bid.
6. If user has highest bid, they can call `claim()` to claim the object. At that point, we decrypt a boolean that shows whether the caller has the highest bid. No other decryptions are done.
7. If user doesn't have the highest bid, they can `withdraw()` their bid.

Currently, we emit an event to denote that the winner has claimed the "object" of the auction.

If two or more bidders bid the same value, whoever claims first is the winner. All other bidders can only withdraw their bids at that point.

### DIFF
--- a/examples/BlindAuction.sol
+++ b/examples/BlindAuction.sol
@@ -70,8 +70,9 @@ contract BlindAuction {
     function withdraw() public {
         require(auctionStopped);
         FHEUInt bidValue = bids[msg.sender];
-        // TODO: Maybe we should now allow a highest bidder to withdraw?
-        Common.requireCt(FHEOps.lte(bidValue, highestBid));
+        if (!objectClaimed) {
+            Common.requireCt(FHEOps.lt(bidValue, highestBid));
+        }
         tokenContract.transfer(msg.sender, bidValue);
         bids[msg.sender] = FHEUInt.wrap(0);
     }
--- a/examples/BlindAuction.sol
+++ b/examples/BlindAuction.sol
@@ -1,0 +1,88 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+pragma solidity >=0.8.13 <0.9.0;
+
+import "../lib/Ciphertext.sol";
+import "../lib/Common.sol";
+import "../lib/FHEOps.sol";
+
+import "./EncryptedERC20.sol";
+
+contract BlindAuction {
+    // The owner of the auction (this) contract.
+    address public auctionOwner;
+
+    // Current highest bid.
+    FHEUInt internal highestBid;
+    // Mapping from bidder to their bid value.
+    mapping(address => FHEUInt) internal bids;
+
+    // The token contract used for encrypted bids.
+    EncryptedERC20 public tokenContract;
+
+    // Whether the auction has stopped. Bids are accepted until it has stopped.
+    bool public auctionStopped;
+
+    // Whether the auction object has been claimed. We need that for a tie-break when
+    // multiple bidders have the same value - in that case, the first one wins.
+    bool public objectClaimed;
+
+    event Winner(address who);
+
+    constructor(EncryptedERC20 _tokenContract) {
+        auctionOwner = msg.sender;
+        tokenContract = _tokenContract;
+        auctionStopped = false;
+        objectClaimed = false;
+    }
+
+    // Bid an `encryptedValue`.
+    function bid(bytes calldata encryptedValue) public {
+        require(!auctionStopped);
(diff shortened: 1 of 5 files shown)


examples/BlindAuction.sol
```diff
@@ -0,0 +1,88 @@
...
+
+    // Bid an `encryptedValue`.
+    function bid(bytes calldata encryptedValue) public {
+        require(!auctionStopped);
+        FHEUInt value = Ciphertext.verify(encryptedValue);
+        bids[msg.sender] = value;
+        if (FHEUInt.unwrap(highestBid) == 0) {
+            highestBid = value;
+        } else {
+            highestBid = FHEOps.cmux(FHEOps.lte(highestBid, value), value, highestBid);
```
immortal-tofu, line 47: Ho I didn't know about this cmux command. It's like a if condition in the end?

examples/BlindAuction.sol
```diff
@@ -0,0 +1,88 @@
...
+        objectClaimed = true;
+        // TODO: For now, just emit an event indicating who won.
+        emit Winner(msg.sender);
+    }
+
+    // Withdraw a bid from the auction to the caller once the auction has stopped.
+    function withdraw() public {
+        require(auctionStopped);
+        FHEUInt bidValue = bids[msg.sender];
+        // TODO: Maybe we should now allow a highest bidder to withdraw?
```
immortal-tofu: `if objectClaimed == true`, we don't need to do any check and transfer directly.

examples/BlindAuction.sol
```diff
@@ -0,0 +1,88 @@
...
+        Common.requireCt(FHEOps.lte(bidValue, highestBid));
```
immortal-tofu: Does it mean that if I've made the highestBid, I can withdraw?

examples/BlindAuction.sol
```diff
@@ -0,0 +1,88 @@
...
+        require(auctionStopped);
+        require(!objectClaimed);
+        Common.requireCt(FHEOps.lte(highestBid, bids[msg.sender]));
+        objectClaimed = true;
+        // TODO: For now, just emit an event indicating who won.
+        emit Winner(msg.sender);
+    }
+
+    // Withdraw a bid from the auction to the caller once the auction has stopped.
+    function withdraw() public {
```
immortal-tofu, line 70: I would see something like this to allow highestBids (because you can have many) to withdraw their money only if object has been claimed. Does it make sense ?
```
    function withdraw() public {
        require(auctionStopped);
        FHEUInt bidValue = bids[msg.sender];
        if (objectClaimed == true) {
            tokenContract.transfer(msg.sender, bidValue);
            bids[msg.sender] = FHEUInt.wrap(0);
        } else {
            Common.requireCt(FHEOps.lt(bidValue, highestBid));
            tokenContract.transfer(msg.sender, bidValue);
            bids[msg.sender] = FHEUInt.wrap(0);
        }
    }
  ```

examples/BlindAuction.sol
dartdart26: Yeah, exactly: `result = cmux(controlBit, ifTrue, ifFalse)`.

examples/BlindAuction.sol
dartdart26: Yes, good catch, will fix it!

examples/BlindAuction.sol
dartdart26: I think yes. One of the reasons is that we haven't yet exposed a less-than or greater-than (without equal). However, I think we should not allow the highest bidder to withdraw, though I am not sure.

examples/BlindAuction.sol
dartdart26: Yes, exactly, we need to get rid of the equality.

dartdart26: Let me actually add `lt` and fix it.

dartdart26: Done using `lt`.

PR #17, opened 2023-03-01
Add support for optimistic ciphertext requires

dartdart26:
Expose the `optimisticRequireCt()` precompile as a library call. See its description for details on how it works.

Add example use of optimistic ciphertext requires with the aim of showing the different gas usage as compared to syncrhonous non-optimistic and plaintext ones.


immortal-tofu (approved):
LGTM

mortendahl (commented):
lgtm

PR #23, opened 2023-03-21
FHEUIntX type boilerplate code

### LINKED ISSUE #20: Generate boilerplate code for Solidity FHEUInt types

tremblaythibaultl:
Introduce a simple python script to generate the FHElib code.

This boilerplate code will let Solidity developer get compile-time type checking with FHEUInt types as well as runtime checking.

Closes #20.


codegen/FHEOps.sol
```diff
@@ -0,0 +1,906 @@
...
+pragma solidity >=0.8.13 <0.9.0;
+
+type euint8 is uint256;
+type euint16 is uint256;
+type euint32 is uint256;
+type euint64 is uint256;
+type euint128 is uint256;
+type euint256 is uint256;
+
+library Common {
```
dartdart26: We have a `Common` library in the Common.sol file. Do these become part of the same library?

codegen/FHEOps.sol
```diff
@@ -0,0 +1,906 @@
...
+    // Values used to communicate types at runtime to the cast() precompile.
```
dartdart26: Got me thinking about whether we should also pass the compile-time types to the runtime in operations too. I think we discussed it, but I forgot what we decided at the end.

codegen/FHEOps.sol
```diff
@@ -0,0 +1,906 @@
...
+
+    function to_euint8(euint16 v) internal view returns (euint8) {
+        return euint8.wrap(Impl.cast(euint16.unwrap(v), Common.euint16_t));
+    }
+
+    function to_euint8(euint32 v) internal view returns (euint8) {
+        return euint8.wrap(Impl.cast(euint32.unwrap(v), Common.euint32_t));
+    }
+
+    function to_euint8(euint64 v) internal view returns (euint8) {
```
dartdart26: Should we call these `toEuint`?

dartdart26:
A thing that comes to mind is that we need to be explicit and document the behaviour on integer overflow. AFAIK, at the moment, we cannot detect it for euint types and make them behave as uint types. Even if we can detect it in the future, we'd have to decrypt a boolean that denotes whether overflow happened. Maybe we could have two versions of these operations - one that checks and one that doesn't.

codegen/FHEOps.sol
tremblaythibaultl: I think we can make a split (one file per library) as we used to.

tremblaythibaultl:
> AFAIK, at the moment, we cannot detect it for euint types and make them behave as uint types. 

To be clear, euint types behave like "regular" `uint`s with wrap-around-on-overflow logic, not solidity uints with revert-on-overflow logic.

tremblaythibaultl:
Closes #20 and is made to be used with https://github.com/zama-ai/go-ethereum/issues/63.

types/TypedEncryptedERC20.sol
```diff
@@ -0,0 +1,119 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+pragma solidity >=0.8.13 <0.9.0;
+
+import "./types/lib/Ciphertext.sol";
+import "./types/lib/Common.sol";
+import "./types/lib/FHEOps.sol";
+
+contract TypedEncryptedERC20 {
```
dartdart26: Should we make all code in the examples folder work with the new library as part of this PR? Say just change from FHEUInt to euint16 or whatever type?

tremblaythibaultl: Sure, I'll add an issue for this.

tremblaythibaultl:
Commit [5788b2d](https://github.com/zama-ai/zbc-solidity/pull/23/commits/5788b2dd252391345177b662c509ffbad577d184) closes #27.

codegen/FHEOps.sol
tremblaythibaultl: Done in [daa4e0b](https://github.com/zama-ai/zbc-solidity/pull/23/commits/daa4e0b67bb405af93713400de789ab5f758c42a).

types/TypedEncryptedERC20.sol
tremblaythibaultl: Done in [5788b2d](https://github.com/zama-ai/zbc-solidity/pull/23/commits/5788b2dd252391345177b662c509ffbad577d184).

PR #28, opened 2023-05-05
Fix reencrypt()'s return value

dartdart26:
Add a missing `reencrypted = new bytes(maxCiphertextBytesLen)` line in the implementation of reencrypt. Not having it leads to a correctly-sized output, but the contents are incorrect - most of the resulting ciphertext serialization bytes are zeroes.


tremblaythibaultl (approved):
lgtm 👍

PR #37, opened 2023-06-01
Clean library and update EIP 712

immortal-tofu:


README.md
```diff
@@ -1,15 +1,31 @@
 # zbc-solidity
+
 A Solidity library for interacting with the Zama Blockchain
 
+# Guide development
```
dartdart26:
```suggestion
# Development Guide
```

README.md
```diff
@@ -1,15 +1,31 @@
...
+
+Install dependencies (Soldiity libraries and dev tools
```
dartdart26:
```suggestion
Install dependencies (Solidity libraries and dev tools)
```

README.md
```diff
@@ -1,15 +1,31 @@
...
+
+```
+npm install
+```
+
+Note: Solidity files are formatted with prettier.
+
 # Demo test
```
dartdart26:
```suggestion
# Demo Test
```

demo_test.py
```diff
@@ -14,9 +14,9 @@
 
 def transfer(contract, to, account, amount):
 	#TODO: use public key encryption instead
```
dartdart26:
```suggestion
```

demo_test.py
```diff
@@ -91,7 +91,7 @@ class Reencrypt(EIP712Struct):
 	print(pt_int)
 	assert pt_int == expected
 
-w3 = Web3(Web3.HTTPProvider('http://localhost:8545', request_kwargs={'timeout': 600}))
+w3 = Web3(Web3.HTTPProvider('http://13.37.31.214:8545', request_kwargs={'timeout': 600}))
```
dartdart26: Should we maybe put a placeholder or comment here such that users change the address and don't connect to our IP?

examples/BlindAuction.sol
```diff
@@ -2,136 +2,129 @@
 
 pragma solidity >=0.8.13 <0.9.0;
 
-import "../lib/Ciphertext.sol";
-import "../lib/Common.sol";
-import "../lib/FHEOps.sol";
+import '../node_modules/@openzeppelin/contracts/utils/cryptography/ECDSA.sol';
+import '../node_modules/@openzeppelin/contracts/utils/cryptography/EIP712.sol';
```
dartdart26: Would it make sense to import directly from GitHub, such that no `node_modules` directory is required? That way users could easily develop in Remix, e.g.:

```
import "https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.9.0/contracts/utils/cryptography/ECDSA.sol"
```

Also for other imports and examples.

examples/BlindAuction.sol
```diff
@@ -2,136 +2,129 @@
...
+  // Mapping from bidder to their bid value.
+  mapping(address => euint8) public bids;
+
+  // The token contract used for encrypted bids.
+  SmallEncryptedERC20 public tokenContract;
+
+  // Whether the auction object has been claimed.
+  bool public objectClaimed;
+
+  // If the token has been transferred to the beneficiary
```
dartdart26, line 28:
```suggestion
  // If the token has been transferred to the beneficiary.
```

examples/BlindAuction.sol
```diff
@@ -2,136 +2,129 @@
...
+  function claim() public onlyAfterEnd {
+    require(!objectClaimed);
+    TFHE.requireCt(TFHE.lte(highestBid, bids[msg.sender]));
+
+    objectClaimed = true;
+    bids[msg.sender] = euint8.wrap(0);
+    emit Winner(msg.sender);
+  }
+
+  // Transfer token to beneficiary
```
dartdart26, line 95:
```suggestion
  // Transfer token to beneficiary.
```

examples/BlindAuction.sol
```diff
@@ -2,136 +2,129 @@
...
+
+  modifier onlyAfterEnd() {
+    if (block.timestamp <= endTime) revert TooEarly(endTime);
+    _;
+  }
+
+  modifier onlySignedPublicKey(bytes memory signature, bytes32 publicKey) {
+    bytes32 digest = _hashTypedDataV4(keccak256(abi.encode(keccak256('Reencrypt(bytes32 publicKey)'), publicKey)));
+    address signer = ECDSA.recover(digest, signature);
+    require(signer == msg.sender, 'EIP712 signer and transaction signer do not match');
```
dartdart26: Would it make sense to put that code for reuse in the library? Say:

```
library TFHE {
  function onlySignedPublicKey(bytes memory signature, bytes32 publicKey) {
    bytes32 digest = _hashTypedDataV4(keccak256(abi.encode(keccak256('Reencrypt(bytes32 publicKey)'), publicKey)));
    address signer = ECDSA.recover(digest, signature);
    require(signer == msg.sender, 'EIP712 signer and transaction signer do not match');
  }
}
```

And then all modifiers become:
```
modifier onlySignedPublicKey(bytes memory signature, bytes32 publicKey) {
  TFHE. onlySignedPublicKey(signature, publicKey);
  _;
}
```

What do you think?

examples/EncryptedERC20.sol
```diff
@@ -2,112 +2,129 @@
...
+import '../lib/TFHE.sol';
+
+contract EncryptedERC20 is EIP712 {
+  euint32 private totalSupply;
+  string public constant name = 'Naraggara'; // City of Zama's battle
+  string public constant symbol = 'NARA';
+  uint8 public constant decimals = 18;
+
+  // used for output authorization
+  bytes32 private DOMAIN_SEPARATOR;
```
dartdart26, line 16:
```suggestion
```

examples/SmallEncryptedERC20.sol
```diff
@@ -0,0 +1,125 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+pragma solidity >=0.8.13 <0.9.0;
+
+import '../node_modules/@openzeppelin/contracts/utils/cryptography/ECDSA.sol';
+import '../node_modules/@openzeppelin/contracts/utils/cryptography/EIP712.sol';
+import '../lib/TFHE.sol';
+
+contract SmallEncryptedERC20 is EIP712 {
```
dartdart26: What is the purpose of this contract?

lib/Common.sol
```diff
@@ -2,51 +2,13 @@
 
 pragma solidity >=0.8.13 <0.9.0;
 
-import "./Precompiles.sol";
-
-// Represents a ciphertext.
-type FHEUInt is uint256;
+type euint8 is uint256;
+type euint16 is uint256;
+type euint32 is uint256;
```
dartdart26, line 7: I guess this whole file and library can now be moved to the TFHE one and we can remove the Common.sol file.

lib/Precompiles.sol
```diff
@@ -2,16 +2,17 @@
...
+    uint256 public constant Delegate = 68;
     uint256 public constant Require = 69;
     uint256 public constant LessThanOrEqual = 70;
     uint256 public constant Subtract = 71;
     uint256 public constant Multiply = 72;
     uint256 public constant LessThan = 73;
-    uint256 public constant Rand = 74;
+    uint256 public constant Random = 74;
     uint256 public constant OptimisticRequire = 75;
+    uint256 public constant Cast = 76;
```
dartdart26:
```suggestion
```

I think it would be better to not expose it till we have it.

lib/Precompiles.sol
```diff
@@ -2,16 +2,17 @@
 
 pragma solidity >=0.8.13 <0.9.0;
 
-// Addresses of precompiled contracts.
 library Precompiles {
     uint256 public constant Add = 65;
     uint256 public constant Verify = 66;
     uint256 public constant Reencrypt = 67;
+    uint256 public constant Delegate = 68;
```
dartdart26:
```suggestion
```

That one is deprecated and we'd better remove it.

lib/Precompiles.sol
```diff
@@ -2,16 +2,17 @@
...
     uint256 public constant Require = 69;
     uint256 public constant LessThanOrEqual = 70;
     uint256 public constant Subtract = 71;
     uint256 public constant Multiply = 72;
     uint256 public constant LessThan = 73;
-    uint256 public constant Rand = 74;
+    uint256 public constant Random = 74;
```
dartdart26:
```suggestion
```

Not available yet, maybe it is better to remove it.

dartdart26 (dismissed):
Looks good!

I've added some minor suggestions, but none are critical.

demo_test.py
mortendahl: putting a placeholder instead makes sense to me too

examples/BlindAuction.sol
```diff
@@ -2,136 +2,129 @@
...
+    endTime = block.timestamp + biddingTime;
+    objectClaimed = false;
+    tokenTransferred = false;
+  }
+
+  // Bid an `encryptedValue`.
+  function bid(bytes calldata encryptedValue) public onlyBeforeEnd {
+    euint8 value = TFHE.asEuint8(encryptedValue);
+    euint8 existingBid = bids[msg.sender];
+    if (euint8.unwrap(existingBid) != 0) {
```
mortendahl, line 56: would it make sense to add a `isZero` or `isNull` function to `euint8` to avoid these `unwrap`? to me they break encapsulation a bit

examples/CMUX.sol
```diff
@@ -2,23 +2,35 @@
...
-        FHEUInt ifTrue = Ciphertext.verify(ifTrueBytes);
-        FHEUInt ifFalse = Ciphertext.verify(ifFalseBytes);
-        result = FHEOps.cmux(control, ifTrue, ifFalse);
-    }
+  constructor() EIP712('Authorization token', '1') {}
 
-    function getResult() public view returns (bytes memory) {
-        return Ciphertext.reencrypt(result);
-    }
+  // Set result = (ifTrue - ifFalse) * control + ifFalse
```
mortendahl:
```suggestion
  // Set result = if control { ifTrue } else { ifFalse }
```

examples/EncryptedERC20.sol
```diff
@@ -2,112 +2,129 @@
...
+    balances[to] = TFHE.add(balances[to], amount);
+    balances[from] = TFHE.sub(balances[from], amount);
+  }
+
+  modifier onlyContractOwner() {
+    require(msg.sender == contractOwner);
+    _;
+  }
+
+  modifier onlySignedPublicKey(bytes memory signature, bytes32 publicKey) {
```
mortendahl: as @dartdart26 suggested, would be nice to have this in a library instead

examples/OptimisticRequire.sol
```diff
@@ -2,80 +2,82 @@
...
+  }
+
+  function getValue() public view returns (uint256) {
+    return value;
+  }
+
+  // Charge full gas as both requires are true.
+  function optimisticRequireCtTrue() public {
+    // True.
+    TFHE.optimisticRequireCt(TFHE.lte(ct1, ct2));
```
mortendahl, line 39: how about we drop the `Ct` suffix? I don't think there's any confusion even for the `TFHE.require` case

demo_test.py
immortal-tofu: I'll leave localhost. This script will heavily change with future @leventdem modifications.

lib/Common.sol
immortal-tofu: Impl.sol needs it, so we can't move it in TFHE.sol or we would have recursive dependencies

examples/BlindAuction.sol
immortal-tofu: Good idea, I'll create an issue to do it later !

examples/EncryptedERC20.sol
immortal-tofu: This is not TFHE, this is EIP712. Should we develop an EIP712 library ?

examples/OptimisticRequire.sol
immortal-tofu: I agree. I create an issue for this.

examples/EncryptedERC20.sol
mortendahl: if it's something that we expect developers will have to do a lot because of our design decisions (which I believe it is?) then yes why not

immortal-tofu: Our blockchain doesn't need EIP-712 to work. `TFHE.reencrypt` reencrypts with the public key provided, signed or not.

In smart contract world, when you want to verify data, you need to ask the user to sign it, and the best way to do it is using EIP-712.

On node/browser side, you can use any EIP-712 libraries such as https://www.npmjs.com/package/eip-712 or https://www.npmjs.com/package/ethers-eip712 for ethers.js. If you want to interact from an other language such as python or Rust, you probably have package also for this.
On solidity you can use OpenZeppelin implementation which is audited and well tested. There are maybe others?

So, based on that, I don't think we should provide and maintain any new library for EIP-712. We are fully compliant with the standard.

```
    bytes32 digest = _hashTypedDataV4(keccak256(abi.encode(keccak256('Reencrypt(bytes32 publicKey)'), publicKey)));
    address signer = ECDSA.recover(digest, signature);
    require(signer == msg.sender, 'Invalid EIP712 signature');
```
Last thing, we don't even need to provide a schema: I decided to choose `Reencrypt` and `publicKey`, but I could use any other name. Maybe at some point we could provide a standard schema if we want to implement specific mechanic like "multicontracts".

immortal-tofu: A trade-off would be to create a custom EIP-712 abstract contract which includes the modifier, and put it in `example` folder. User could copy/paste it if they want it.

```
import '../node_modules/@openzeppelin/contracts/utils/cryptography/ECDSA.sol';
import '../node_modules/@openzeppelin/contracts/utils/cryptography/EIP712.sol';

abstract contract EIP712 is EIP712 {
  modifier onlySignedPublicKey(bytes memory signature, bytes32 publicKey) {
    bytes32 digest = _hashTypedDataV4(keccak256(abi.encode(keccak256('Reencrypt(bytes32 publicKey)'), publicKey)));
    address signer = ECDSA.recover(digest, signature);
    require(signer == msg.sender, 'Invalid EIP712 signature');
    _;
  }
}

immortal-tofu: Just added the abstract https://github.com/zama-ai/zbc-solidity/pull/37/commits/789631b4615921d58cf2f88bb12fad33046f1777#diff-d4bc17573850490edf605ada302acfb122069860195241d7e7bfd1bf0a877de8

mortendahl: could it simply be a helper function that we offer? even if users have a choice to do things differently, if we believe this is best practice then why not make life a bit easier for them (and avoid too much copy/paste)

immortal-tofu: We plan to ship our library as npm package (it seems it's the standard for Solidity contracts). Today we have 0 dependency.

If we decide to ship a "helper" for OpenZeppelin, it would mean adding the [OpenZeppelin](https://www.npmjs.com/package/@openzeppelin/contracts) package in our package.json (+1 dependency), with potential version conflicts for the user (if we decide to import OpenZeppelin x.x.x and user's project uses a different version). I'm not fan of adding an extra dependency to create a 3 lines helper. I prefer have a clear documentation and examples for this.

That's my point of view but if you prefer to add this dependency and move this abstract contract in the lib folder I can do it.

immortal-tofu: However, if at some point we decide to create a "multiple contracts" token, we'll need to implement our own custom usage of EIP-712 and ship it as an helper.

dartdart26 (approved):

PR #41, opened 2023-06-13
Remove the cast and delegate precompiles

dartdart26:
Rationale is that cast is not yet supported and delegate is obsolete.

Furthermore, add a few minor style fixes.

### DIFF
--- a/codegen/codegen.py
+++ b/codegen/codegen.py
@@ -11,7 +11,7 @@ type euint16 is uint256;
 type euint32 is uint256;
 
 library Common {
-// Values used to communicate types at runtime to the cast() precompile.
+    // Values used to communicate types to the runtime.
     uint8 internal constant euint8_t = 0;
     uint8 internal constant euint16_t = 1;
     uint8 internal constant euint32_t = 2;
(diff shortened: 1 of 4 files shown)


immortal-tofu:
Did you use prettier ? We don't check (yet) about formatting in CI.
If not, just `npm install` and you can use `npm run prettier`. Otherwise, all code editor have a prettier plugin to autoformat on save.

immortal-tofu (approved):

dartdart26:
> Did you use prettier ? We don't check (yet) about formatting in CI.
> If not, just `npm install` and you can use `npm run prettier`. Otherwise, all code editor have a prettier plugin to autoformat on save.

I generated the code with the codegen script. Maybe we could run prettier as part of code generation?

dartdart26:
> > Did you use prettier ? We don't check (yet) about formatting in CI.
> > If not, just `npm install` and you can use `npm run prettier`. Otherwise, all code editor have a prettier plugin to autoformat on save.
> 
> I generated the code with the codegen script. Maybe we could run prettier as part of code generation?

Maybe we could do it in Python itself. Let me know if you think we should merge as is and deal with format in a separate PR or do it here.

Issue #46, opened 2023-06-20
FHE operations with literals


mortendahl:
Add syntactic sugar for doing FHE operations involving literals, say `FHE.add(x, 5)`, by inlining the trivial encryption call

PR #48, opened 2023-06-21
feat(precompile): add trivial encrypt

tremblaythibaultl:
Add an overload for `asEuintX` that takes as input a `uint256` value and returns a handle to the trivial encryption of said value.
Contributes to zama-ai/go-ethereum#116 and #46.
Meant to be used with zama-ai/go-ethereum/pull/117.


dartdart26 (approved):
Looks good!

As a side note, maybe we need to somehow ensure a prettier is run, because we don't have linters in the Python comment. I think there is a call to `npm prettier` in package.json.

As another side note, I am wondering if we can get away without npm and JS in the Solidity library, but I think @immortal-tofu knows more.

PR #50, opened 2023-06-21
feat: add casting

tremblaythibaultl:
Add support for casting. Meant to be used with zama-ai/go-ethereum#118. Closes zama-ai/go-ethereum#119.


dartdart26 (dismissed):
Looks good!

I think we need to somehow make prettier automatic or use it in our IDEs.

dartdart26 (approved):

PR #57, opened 2023-06-26
Update EncryptedERC20.sol

tremblaythibaultl:
Remove the comment indicating that a user's balance is re-encrypted under its FHE public key, which is automatically determined (this is not true anymore).


immortal-tofu (approved):

immortal-tofu:
lgtm

PR #60, opened 2023-06-26
Update of library and blind auction

immortal-tofu:
- Add isInitialized method
- Rename requireCt/optimisticRequireCt to req and optReq
- Update Blind Auction contract
- Fix issue with code duplication on codegen generation


tremblaythibaultl (approved):
lgtm 👍

PR #61, opened 2023-06-27
package() prepare repository for npm package

immortal-tofu:
- Updated README and package.json to publish on NPM
- Updated paths for correct import in VSCode or Remix IDE
- I change the demo_test.py to allow import: it seems to work locally, but it would be nice to check the full chain with the modification. It seems the docker image doesn’t work on M1, so I just tested locally by executing the python script. @leventdem

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,5 @@
+*
+!lib/*
+!package.json
+!README.md
+!LICENSE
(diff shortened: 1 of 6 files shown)


tremblaythibaultl (approved):
looks good to me :)

PR #62, opened 2023-06-28
Add scalar ops

### LINKED ISSUE #58: Use actual ciphertext/plaintext ops

tremblaythibaultl:
Add scalar FHE ops that use the actual scalar op C bindings exposed from TFHE-rs. 
Closes #58.
Contributes to zama-ai/go-ethereum#116.

The lack of a test suite for our solidity library and the amount of changes this PR includes means it should be thoroughly tested as soon as we can.


immortal-tofu:
LGTM but we clearly need to implement test on all operations

immortal-tofu (approved):

PR #64, opened 2023-06-29
fix(impl): fix mul bug

tremblaythibaultl:
Fix bug where `mul` would return a wrong result in the case where one of the two operands is `0`.


dartdart26 (approved):
Looks good!

PR #65, opened 2023-06-30
Add scalar ops for same size operators

tremblaythibaultl:
And fix a bug affecting comparison scalar operators.


dartdart26 (approved):
Looks good!

PR #72, opened 2023-07-07
fix(tfhe): add automatic initialization for ciphertexts

tremblaythibaultl:
This PR removes the following incorrect logic of some binary operators:
```solidity
if (a == 0) {
  return b;
} else if (b == 0) {
  return a;
}
```
and introduces initialization logic such that if a `0` handle is passed to any binary operator, the handle is set to point to the trivial encryption of `0`.
This makes it so that FHE operations on uninitialized ciphertexts do not revert, even in the case of ciphertext/plaintext operations.

This PR also introduces a fix for a bug regarding ciphertext/plaintext operations and Solidity's [overload resolution](https://docs.soliditylang.org/en/v0.8.20/contracts.html#overload-resolution-and-argument-matching).
With this fix, ciphertext/plaintext operations are only defined for values of the same bit width, meaning `add(euint32 a, uint32 b)` is defined but `add(euint16 a, uint32 b)` is not.
It is worth noting that the overload resolution process will implicitely upcast the `uint` value such that both
```solidity
add(euint16 a, uint8 b)
add(euint16 a, uint16 b)
```
will work but
```solidity
add(euint16 a, uint32 b)
```
will not.

### DIFF
--- a/codegen/codegen.py
+++ b/codegen/codegen.py
@@ -75,12 +75,6 @@ library Impl {
     uint256 constant fhePubKeySize = 32 + 16553;
 
     function add(uint256 a, uint256 b, bool scalar) internal view returns (uint256 result) {
-        if (a == 0) {
-            return b;
-        } else if (b == 0) {
-            return a;
-        }
-        
         bytes1 scalarByte;
         if (scalar) {
             scalarByte = 0x01;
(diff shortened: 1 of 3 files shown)


dartdart26 (approved):
Looks good!

PR #73, opened 2023-07-07
Abstract in package + bugfix contractowner blind auction

immortal-tofu:
- Merge the fix on blinc auction for stop()
- Move abstract at root and add it in the npm package.
- Move @openzeppelin dependency as non-dev dependency


tremblaythibaultl (approved):
lgtm 👍

PR #76, opened 2023-07-10
fix(tfhe): add `neg` and `not` overloads

tremblaythibaultl:
Add overloads for the `neg` and `not` unary operators.


dartdart26 (approved):
Looks good!

PR #84, opened 2023-07-12
fix(auction): remove use of `euint32.wrap(0)`

### LINKED ISSUE #77: Use `delete` in example contracts
Some smart contracts use `euint32.wrap(0)` to delete entries in mappings, but this is leaking implementation details, and it would be better to use the `delete` keyword

tremblaythibaultl:
Introduces `TFHE.NIL` constants to provide initial values for `euint` types.

Note: we cannot use the Solidity `delete` operator to set a variable to its initial value because the `delete` operator is not defined for user-defined value types (e.g., `euint32`). 

This will change in the future when we move to a higher version of the Solidity compiler.

Closes #77.

### DIFF
--- a/codegen/codegen.py
+++ b/codegen/codegen.py
@@ -9,6 +9,9 @@ pragma solidity >=0.8.13 <0.8.20;
 type euint8 is uint256;
 type euint16 is uint256;
 type euint32 is uint256;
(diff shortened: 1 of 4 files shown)


dartdart26 (approved):
Nice!

PR #85, opened 2023-07-12
Add method documentation in TFHE.sol

dartdart26:
Add documentation for methods that didn't have any. Move doc that was in Impl.sol to TFHE.sol (e.g. cmux, optReq).

Also, fix neg() and not() in TFHE.sol such that they return a result.


dartdart26:
Resolves #85

tremblaythibaultl (approved):
lgtm 👍

PR #86, opened 2023-07-13
feat(tfhe): add support for `ebool` type

tremblaythibaultl:
Add support for the `ebool` type. Acts as another wrapper over an `FheUint8`. 
Sort of fixes zama-ai/go-ethereum#127, in the sense that the real `ebool` type is not available yet and this is only syntactic sugar.

### DIFF
--- a/codegen/codegen.py
+++ b/codegen/codegen.py
@@ -1003,15 +1003,15 @@ to_print="""
 
 to_print_cast="""
     // Cast an encrypted integer from euint{i} to ebool.
-    function asEbool(euint{i} ciphertext) internal view returns (ebool) {{
-        return ne(ciphertext, 0);
+    function asEbool(euint{i} value) internal view returns (ebool) {{
+        return ne(value, 0);
     }}
 """
 
 to_print_no_cast="""
     // Cast an encrypted integer from euint8 to ebool.
-    function asEbool(euint8 ciphertext) internal view returns (ebool) {{
-        return ne(ciphertext, 0);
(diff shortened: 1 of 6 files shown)


codegen/codegen.py
```diff
@@ -894,10 +1001,42 @@
     }}
 """
 
+to_print_cast="""
+    // Cast an encrypted integer from euint{i} to ebool.
+    function asEbool(euint{i} ciphertext) internal view returns (ebool) {{
```
dartdart26: Nit: I've changed the naming in my previous commit to use `value` when we refer to euint/ebool variables and `ciphertext` when we refer to a serialized (bytes) ciphertext. Maybe we could use `value` here and in other cases too.

codegen/codegen.py
```diff
@@ -852,34 +950,43 @@
...
-    f.write(to_print_scalar.format(i=i, f="gt", g="lt"))
-    f.write(to_print_scalar.format(i=i, f="le", g="ge"))
-    f.write(to_print_scalar.format(i=i, f="lt", g="gt"))
     f.write(to_print_scalar.format(i=i, f="min", g="min"))
     f.write(to_print_scalar.format(i=i, f="max", g="max"))
 
 
 to_print =  """
     // If `control`'s value is 1, the result has the same value as `a`.
     // If `control`'s value is 0, the result has the same value as `b`.
```
dartdart26, line 987: We could change the comment to say true and false instead of 1 and 2.

dartdart26 (approved):
Looks good!

PR #89, opened 2023-07-21
Add support for TFHE.decrypt()

dartdart26:
The decrypt() function decrypts the given euint value. There are a few
considerations when doing decryption:
 * if a transaction is reverted due to a failed TFHE.optimisticReq(),
   then any decryptions would leak information as they would already
   have been decrypted when the txn is reverted at the end (due to the
   nature of TFHE.optimistiReq())
 * gas estimation might be more imprecise, because control flow can now
   depend on decrypted values (that gas estimation doesn't know of as it
   always uses 0)

### DIFF
--- a/codegen/codegen.py
+++ b/codegen/codegen.py
@@ -1002,7 +1002,7 @@ for i in (2**p for p in range(3, 6)):
     f.write(to_print_scalar.format(i=i, f="max", g="max"))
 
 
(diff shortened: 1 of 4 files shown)


immortal-tofu (approved):
LGTM!

tremblaythibaultl (approved):
lgtm 👍

PR #106, opened 2023-08-11
Add div scalar function

david-zk:
Added div call to solidity library.

I added a new template for printing `to_print_scalar_only` which is only for scalar operators, because other templates used `scalar` boolean flag which is not needed here because we only know it will be scalar (I saw other templates `to_print_no_scalar_...` do the same)

@tremblaythibaultl


tremblaythibaultl:
To be used with zama-ai/go-ethereum#151.

tremblaythibaultl (commented):
lgtm 👍

tremblaythibaultl (approved):

PR #109, opened 2023-08-22
Draft test suite for TFHE operations

david-zk:
This is how tests could look, I imagine contract code could be generated with python as it is generated for all the TFHE library functions not to forget any overloads

@tremblaythibaultl

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,12 @@
     "prettier": "prettier --write \"**/*.{js,json,md,sol,ts,yml}\"",
     "test": "hardhat test",
     "typechain": "cross-env TS_NODE_TRANSPILE_ONLY=true hardhat typechain",
-    "codegen": "cd ./lib && python ../codegen/codegen.py && npm run prettier && cd .."
+    "codegen": "cd ./lib && python ../codegen/codegen.py && npm run prettier && cd ..",
+    "task:getEthereumAddress": "hardhat task:getEthereumAddress",
+    "fhevm:start": "docker run -i -p 8545:8545 --rm --name fhevm ghcr.io/zama-ai/evmos-dev-node:v0.1.7",
+    "fhevm:stop": "docker rm -f fhevm",
+    "fhevm:restart": "fhevm:stop && fhevm:start",
+    "fhevm:faucet": "docker exec -i fhevm faucet $(npx hardhat task:getEthereumAddress)"
   },
   "repository": {
     "type": "git",
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,4 +19,11 @@ jobs:
           node-version: ${{ matrix.node-version }}
       - run: cp .env.example .env
       - run: npm ci
-      - run: npm test
+      - name: "npm CI test"
+        run: |
+          npm run fhevm:start &> fhevm.log &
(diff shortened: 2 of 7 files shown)


tremblaythibaultl:
Looks good 👍 
I agree tests can be generated with python at first once and then manually added when new features are added.

tremblaythibaultl:
cc @immortal-tofu

package.json
```diff
@@ -18,7 +18,12 @@
     "prettier": "prettier --write \"**/*.{js,json,md,sol,ts,yml}\"",
     "test": "hardhat test",
     "typechain": "cross-env TS_NODE_TRANSPILE_ONLY=true hardhat typechain",
-    "codegen": "cd ./lib && python ../codegen/codegen.py && npm run prettier && cd .."
+    "codegen": "cd ./lib && python ../codegen/codegen.py && npm run prettier && cd ..",
+    "task:getEthereumAddress": "hardhat task:getEthereumAddress",
+    "fhevm:start": "docker run -i -p 8545:8545 --rm --name fhevm ghcr.io/zama-ai/evmos-dev-node:v0.1.7",
```
immortal-tofu, line 23: These scripts already exist, you merged @leventdem PR yesterday with these scripts, no?

david-zk: Yeah, I literally copied this from template repo 😊

immortal-tofu: Ah right, the PR was on hardhat-template, not this repo. My bad

immortal-tofu (approved):

package.json
david-zk: I'm not sure how I could reuse the template repo though, I don't have much node experience, so I pasted same implementations here 🤔

immortal-tofu:
Looks good. I'd write a folder with operations:
```
test/operation/add.ts
test/operation/mul.ts
...
```
And same for fixture and sol contract. I think it will be easier to maintain by keeping everything atomic. wdyt?

david-zk:
> Looks good. I'd write a folder with operations:
> 
> ```
> test/operation/add.ts
> test/operation/mul.ts
> ...
> ```
> 
> And same for fixture and sol contract. I think it will be easier to maintain by keeping everything atomic. wdyt?

Hmmm I think it would be very many files, and in go ethereum all these operations are defined in single file, same in the solidity library, also would probably take more time to test when deploying separate contracts

immortal-tofu:
For python generation, I wouldn't do this. There is probably a way to do it with hardhat?

david-zk:
> For python generation, I wouldn't do this. There is probably a way to do it with hardhat?

I'd use python likely as one off script that would never end up in this repo, just generate for all the operations so far, other operation contributors would need to add tests manually then

immortal-tofu:
> Hmmm I think it would be very many files, and in go ethereum all these operations are defined in single file, same in the solidity library, also would probably take more time to test when deploying separate contracts

You're right, but we can deploy one contract for all these tests. So it would mean having one .sol contract (maybe generated at some point by codegen.py) and then multiple test suite using the same deployed contract.

The idea is to avoid the `codegen.py` syndrom (= having a 1200 lines unreadable file)

david-zk:
Okay so I merge and in separate PR I'll add the rest of the tests

PR #114, opened 2023-08-28
Remove codegen sorting code

david-zk:
This just removes ordering code which kept codegen in the same as legacy `codegen.py`, diff for TFHE.sol is large but semantic meaning doesn't change.

@immortal-tofu @tremblaythibaultl

### DIFF
--- a/codegen/common.ts
+++ b/codegen/common.ts
@@ -8,7 +8,6 @@ export type Operator = {
   hasEncrypted: boolean;
   arguments: OperatorArguments;
   returnType: ReturnType;
-  tfheSolOrder: number;
 };
 
 export type Precompile = {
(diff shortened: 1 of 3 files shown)


immortal-tofu:
Maybe write more tests on operators before merging this, or we are 100% sure that order won't affect any edge case ?

david-zk:
> Maybe write more tests on operators before merging this, or we are 100% sure that order won't affect any edge case ?

I planned to work on all tests on the next PR, I'm 1000% sure that changing order of methods in solidity file will not affect the logic of the library

david-zk:
@immortal-tofu https://ethereum.stackexchange.com/questions/66541/does-the-ordering-of-fields-in-a-contract-matter#:~:text=In%20general%2C%20the%20compiled%20code,your%20fields%20or%20your%20functions.
> In general, the compiled code is not affected by the ordering of your fields or your functions

immortal-tofu (approved):
Since we are doing a lot of function overload, I wanted to be sure that order won't affect anything. I also read few articles and it seems that there is no issue, so LGTM 👍

PR #115, opened 2023-08-29
Contracts() Governor DAO

immortal-tofu:


tremblaythibaultl:
Good work! 
Do you think it's worth adding tests that check that a winning proposal can get executed properly?

immortal-tofu:
Yes clearly, I'll add that 👍

tremblaythibaultl (approved):
lgtm 👍

PR #116, opened 2023-08-31
Add binary operator tests in solidity

david-zk:
481 operator tests added.
This is enforced with asserts, binary operator cannot exist without tests or build fails.

Operator changes:
- remove scalar div operator where scalar is left
- perform trivial encryption in scalar for sub/shl/shr operators

@tremblaythibaultl @immortal-tofu @dartdart26


codegen/common.ts
```diff
@@ -1,13 +1,22 @@
...
+  // swapped
   leftScalarInvertOp?: string;
   precompileName: string;
   hasScalar: boolean;
   hasEncrypted: boolean;
   arguments: OperatorArguments;
   returnType: ReturnType;
+  // if true do trivial encryption
+  // for left scalar operand, this is workaround
+  // until tfhe-rs supports left scalar operands
```
dartdart26, line 16: Nit: Putting these comments on a single line of their own would aid in readability.

dartdart26: Nit: Putting these comments on a single line of their own would aid in readability.

codegen/templates.ts
```diff
@@ -207,17 +229,34 @@ function tfheScalarOperator(lhsBits: number, rhsBits: number, operator: Operator
...
-  var implExpressionB = `Impl.${operator.name}(euint${outputBits}.unwrap(b), uint256(a)${scalarFlag})`;
-  if (operator.leftScalarInvertOp) {
-    implExpressionB = `Impl.${operator.leftScalarInvertOp}(euint${outputBits}.unwrap(b), uint256(a)${scalarFlag})`;
+  var implExpressionB = `Impl.${leftOpName}(euint${outputBits}.unwrap(b), uint256(a)${scalarFlag})`;
+  var maybeEncryptLeft = '';
+  if (operator.leftScalarEncrypt) {
+    // workaround until tfhe-rs left scalar support:
+    // do the trivial encryption and preserve order of operations
+    scalarFlag = ', false';
+    maybeEncryptLeft = `euint${outputBits} a_enc = asEuint${outputBits}(a);`;
```
dartdart26: Nit: I think we use camelCase in Solidity, i.e. maybe `aEnc`.

dartdart26:
Do you think we should move the generated test contracts to `contracts/tests` instead of directly putting them in `contracts`?

contracts/TFHETestSuite2.sol
```diff
@@ -0,0 +1,531 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+pragma solidity >=0.8.13 <0.8.20;
+
+import "../lib/TFHE.sol";
+
+contract TFHETestSuite2 {
+    function add_euint32_euint8(uint32 a, uint8 b) public view returns (uint32) {
+        euint32 a_proc = TFHE.asEuint32(a);
+        euint8 b_proc = TFHE.asEuint8(b);
```
dartdart26: Nit: Rename these to `aEnc/bEnc`?

test/tfheOperations/tfheOperations.ts
```diff
@@ -14,20 +37,2421 @@ describe('TFHE operations', function () {
...
+    this.contract2Address = await contract2.getAddress();
+    this.contract2 = contract2;
+    const instances2 = await createInstances(this.contract2Address, ethers, this.signers);
+    this.instances2 = instances2;
+  });
+
+  it('test operator "add" overload (euint8, euint8) => euint8 test 1 (3, 4)', async function () {
+    const res = await this.contract1.add_euint8_euint8(3, 4);
+    expect(res).to.equal(7);
+  });
```
dartdart26, line 51: Are these blocks auto-generated?

dartdart26 (approved):
Looks good!

I think what would be useful is to add info on how new operators can be added, e.g. where to add them, where to add tests, which files are auto-generated, etc.

david-zk:
> Do you think we should move the generated test contracts to `contracts/tests` instead of directly putting them in `contracts`?

I see no problem with that, will change

test/tfheOperations/tfheOperations.ts
david-zk: Yes they are

david-zk:
@dartdart26 addressed all comments in 9884ed36bb52a23cbe63d67bbdadf500da8a67fe commit

PR #118, opened 2023-08-31
test() add utils to test with signers

immortal-tofu:


leventdem (approved):
Nice simplifications.  
+1

PR #121, opened 2023-09-05
Remove `TFHE.req`

tremblaythibaultl:
Removes the encrypted require functionality. Instead, we encourage users to use `TFHE.decrypt` together with a regular Solidity `require` statement
To be used with zama-ai/go-ethereum/pull/158.


leventdem (approved):
+1

PR #123, opened 2023-09-06
Add support for the FHE rand precompile

dartdart26:
At the moment, the underlying implementation is using a plaintext PRNG. Because of that, we put a warning for users to be aware of.

Support generation of 8, 16 and 32-bit unsigned integers. Users call it by `TFHE.randEuint32()` and so on for other types. Can only be called in transactions - calling it in view functions will fail.

Add tests.

Finally, update the `fhevmjs` dependency to `~0.2.0`.


tremblaythibaultl (approved):
lgtm 👍

PR #124, opened 2023-09-06
Chore/add tests for blind auction

leventdem:


.github/workflows/test.yml
```diff
@@ -27,5 +27,8 @@ jobs:
           tail -f fhevm.log | sed '/Starting JSON WebSocket server/ q'
 
           npm run fhevm:faucet
+          npm run fhevm:faucetBob
```
immortal-tofu, line 30: A possible evolution would be to wait for transaction validation before exit the command

codegen/templates.ts
```diff
@@ -692,20 +687,6 @@ function implCustomMethods(): string {
         result = uint256(output[0]);
     }
 
-    function req(uint256 ciphertext) internal view {
```
immortal-tofu, line 695: This is not something done by @tremblaythibaultl in an other PR ?

lib/Impl.sol
```diff
@@ -580,20 +580,6 @@ library Impl {
         result = uint256(output[0]);
     }
 
-    function req(uint256 ciphertext) internal view {
```
immortal-tofu, line 583: Same comment

immortal-tofu (approved):

immortal-tofu:
Few comments but otherwise LGTM

codegen/templates.ts
leventdem: yes just because of a rebase

.github/workflows/test.yml
leventdem: if we wait in between it is generally ok

PR #132, opened 2023-09-14
Fix codegen for tests

david-zk:
Now test code generated contracts are out of sync

@dartdart26 @immortal-tofu


dartdart26 (approved):
Looks good!

Should we put `examples/test/*` or `examples/test/TFHETestSuite*` to .gitnignore? Or commit the generated files?

PR #135, opened 2023-09-18
Rework overload tests not to use trivial encryption

david-zk:
Reworked operator tests to not use trivial encryption, also one more test shard because when testing two contract shards `bytes calldata` is too big to deploy

@leventdem @immortal-tofu


leventdem (approved):
+1

PR #144, opened 2023-09-19
Use fheLib precompile for all internal functions

david-zk:
Move all of our library functions to use single fheLib precompiled smart contract through interface underneath

@immortal-tofu @tremblaythibaultl @dartdart26 

@dartdart26 about function purity, I think one strategy could be to have two interfaces, one with pure functions for overloads and another interface for backwards compatibility marked as pure. That way, only operators assume purity and we could introduce operator overloading as experimental feature.


tremblaythibaultl (approved):
lgtm 👍

PR #146, opened 2023-09-20
Operator overloading

david-zk:
Depends on https://github.com/zama-ai/fhevm/pull/144

following operators are supported:
+, - (unary and binary), *, &, |, ^, ~ (unary)

==, !=, <, <=, >, >= cannot be implemented because they must return type of bool where we return ebool.

@dartdart26 @immortal-tofu @tremblaythibaultl


codegen/common.ts
```diff
@@ -96,6 +102,7 @@ export const ALL_OPERATORS: Operator[] = [
     arguments: OperatorArguments.Binary,
     returnType: ReturnType.Uint,
     leftScalarDisable: true,
+    binarySolidityOperator: '*',
```
tremblaythibaultl: Did you mean `/` ?

david-zk: Yeah, leftover, division has no operator because we have only scalar division and all operator inputs must be same types

leventdem (approved):
+1

PR #170, opened 2023-09-25
Add doc about random encrypted integers

dartdart26:
Add doc about the randEuintX() family of functions.

Fix a comment about mixing plaintext and encrypted operands. Remove duplicate sentence about different operand types.


tremblaythibaultl (approved):
Good catches 👍

immortal-tofu (approved):

dartdart26:
Will open a separate one where I put info on what's coming in terms of features too.

PR #171, opened 2023-09-25
fix() force 0.8.19 version for everything

immortal-tofu:
Without this fix, if a contract is written with a solidity version below, it raises an error because overload is not supported


leventdem (approved):
+1

PR #176, opened 2023-09-26
docs() change architecture of Solidity doc

immortal-tofu:
Review here https://app.gitbook.com/s/TTmWgFOo4yfgdZUDHVfj/~/revisions/H93fWbF86XOPFKtjHinl/


docs/solidity/condition.md
```diff
@@ -0,0 +1,96 @@
...
+
+  // Replace highest bid
+  highestBid = bid;
+}
+```
+
+In this code, we first evaluate a homomorphic comparison checking that the user has bid more than the highest bid. This homomorphic comparison will return an encryption of 0 if false, or an encryption of 1 if true. Since we are decrypting this value with `TFHE.decrypt`, we are leaking information: if the user didn't bid enough tokens, the transaction is reverted.
+For example, a user can know the value of the highest bid by trying every possible values and finally bid just one token above.
+
+## Condition homomorphically
```
leventdem: Is Homomorphic condition better ?

docs/solidity/condition.md
```diff
@@ -0,0 +1,96 @@
+# Condition
+
+The result of [comparison operations](functions.md#comparison-operation-eq-ne-ge-gt-le-lt) is of type `ebool`. Typical boolean operations are not currently supported for this type.
+
+That said, there are possibilities to do condition with or without leaking informations.
```
leventdem: It is like information leakage is more suitable.

docs/solidity/decrypt.md
```diff
@@ -0,0 +1,70 @@
+# Decrypt and reencrypt
+
+## Decrypt
+
+We allow explicit decryption requests for any encrypted type.
+The values are decrypted through the distributed decryption protocol and are stored on-chain.
```
leventdem: It is a bit early for now, can we say in a short future we will use ...

docs/solidity/decrypt.md
```diff
@@ -0,0 +1,70 @@
...
+function balanceOf(bytes32 publicKey) public view returns (bytes memory) {
+  return TFHE.reencrypt(balances[msg.sender], publicKey, 0);
+}
+```
+
+> **_NOTE:_** If one of the following operations is called with an uninitialized ciphertext handle as an operand, this handle will be made to point to a trivial encryption of `0` before the operation is executed.
+
+### Handle private reencryption
+
+In the example above (`balanceOf`), this view function need to validate the user to prevent anyone to reencrypt any user's balance. To prevent this, the user provides a signature of the provided public key. The best way to do it is to use [EIP-712 standard](https://eips.ethereum.org/EIPS/eip-712). Since this is something very useful, fhEVM library provide an abstract to use in your contract:
```
leventdem: to avoid repetition, just replace provided by "given".

leventdem (commented):
Proposed some modifications.

leventdem (approved):
+1

PR #185, opened 2023-09-30
feat() add option to run in parallel

immortal-tofu:
- Run in parallel reduce time by 2x (14min => 7min) on Macbook pro M1
- Run in band for CI (sequential) because there is no improvement (monocore)

How it works
- To avoid issue with nonce, it generates 4 random wallets for every test and do the faucet command directly in the test.
- In the `before` method, it's needed to define how many wallet we want to use to avoid useless call to faucet (e.g `initSigners(3)` to give coins to alice, bob and carol)

Typical before method for tests:
```
  before(async function () {
    await initSigners(3);
    this.signers = await getSigners();
  });
```

Improvement:
- Increase default timeout to 180s for all function (before and beforeEach included)

To improve:
- We do a syscall to execute faucet. It would be better to emit a transaction from the test directly
- We don't check if network is local or not. We can do that by reading `config`.


leventdem (approved):
Nice way to prevent nonce issue by generating random wallet.

PR #192, opened 2023-10-05
ci: accounts faucet is automatically handled in tests (no longer need…

leventdem:
… to call faucet manually on premise)


immortal-tofu (approved):

immortal-tofu:
You can use `npm run test:inband` instead: it will run test in band (vs parallel). CI is monothread so let's keep in band test.

See https://github.com/zama-ai/fhevm/blob/f45a320a302fbb0ff3fec4736365a1ffba448201/.github/workflows/test.yml#L29

PR #194, opened 2023-10-05
docs() clarify roadmap section

immortal-tofu:
I moved roadmap in a specific page because now, you need to scroll horizontally to see if it's supported or not.


dartdart26 (approved):
Nice!

PR #195, opened 2023-10-06
Fix optimistic requires

david-zk:
Call optimistic require precompile using assembly instead of fhelib interface.

This is needed because if we call through interface solidity compiler generates `EXTCODESIZE` and checks that our precompile code size is not zero, which it is, even though it is code.

Here is where code is generated in solidity compile for the check that makes our transactions to revert https://github.com/ethereum/solidity/blob/b54e7207739e48c5a7712a4683a9591a6c64b7cd/libsolidity/codegen/ExpressionCompiler.cpp#L2784

@dartdart26 @tremblaythibaultl @immortal-tofu

### DIFF
--- a/codegen/templates.ts
+++ b/codegen/templates.ts
@@ -126,7 +126,6 @@ function generateImplFhevmLibInterface(operators: Operator[]): string {
 
 function fheLibCustomInterfaceFunctions(): string {
   return `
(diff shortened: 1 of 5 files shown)


dartdart26 (approved):
Great work!

tremblaythibaultl (approved):
lgtm 👍

PR #196, opened 2023-10-06
fix(tests): get container name from env instead of hardcoded

youben11:


leventdem (approved):
Nice!  I will update the call in evmos with TEST_CONTAINER_NAME=evmosnodelocal0

PR #202, opened 2023-10-16
test() gas can't be above 10000000

immortal-tofu:


leventdem (approved):
+1
Default value on Metamask is also 1.2  I think.

PR #209, opened 2023-10-23
Add network1 development support

david-zk:
Run one codegen to adapt TFHE library for network1, non-breaking. Some tests already passing. Tweaked the docs @immortal-tofu.

Remaining open questions:
How to adjust tests for the platforms which have operations disabled, say, disabled optimistic and synchronous requires for network1
@dartdart26 @youben11


docs/client/examples/getbalance.md
```diff
@@ -23,9 +23,14 @@ const getInstance = async () => {
   const chainId = +network.chainId.toString();
 
   // Get blockchain public key
-  const publicKey = await provider.call({
-    to: "0x0000000000000000000000000000000000000044",
+  const ret = await provider.call({
+    // fhe lib address
+    to: "0x000000000000000000000000000000000000005d",
```
dartdart26, line 28: Nit: Should we add a hint in the comment telling users to change the `to` address based on the network they are connecting to?

dartdart26 (approved):
Looks good to me!

docs/client/examples/getbalance.md
david-zk: Ok I'll add

PR #211, opened 2023-10-26
feat() Add method for ebool (not,and,or,cmux)

immortal-tofu:


codegen/templates.ts
```diff
@@ -452,6 +452,23 @@ function tfheAsEboolUnaryCast(bits: number): string {
     function asEuint8(ebool b) internal pure returns (euint8) {
         return euint8.wrap(ebool.unwrap(b));
     }
+
+    // Implement and/or operator
```
dartdart26: To be consistent with comments on other functions, might add comment for all functions, e.g. `Evaluate and(a, b) and return the result.` and so on for `or` and `not`.

For `cmux`, could be:

```
    // If 'control''s value is 'true', the result has the same value as 'a'.
    // If 'control''s value is 'false', the result has the same value as 'b'.
```

dartdart26 (approved):
Looks good! Added a minor comment.

PR #213, opened 2023-11-02
Fix network1 tests waiting for new block

david-zk:
Adds utility function to generate transactions so blocks would be always produced by network1 (there blocks are not produced unless transactions are added)

Also adds new eve signer account (if the same ones reused the utility function to generate transactions don't work)

@immortal-tofu @dartdart26


dartdart26 (approved):
Looks good! Thanks for fixing it!

immortal-tofu:
Good catch!

immortal-tofu (approved):

PR #214, opened 2023-11-02
fix() stateful optimistic require false test

dartdart26:
Instead of expecting an exception, expect that the receipt's status will be different than 1, i.e. failing.


david-zk (approved):

dartdart26:
I am confused as to why I get different behaviour on my machine...

dartdart26:
Doing `npm install` fixed it on my machine.

PR #216, opened 2023-11-06
Increase timeouts for identified erc20 tests

david-zk:
Tests fail now with smaller timeout

@dartdart26 @immortal-tofu @youben11

### DIFF
--- a/test/governor/GovernorZama.ts
+++ b/test/governor/GovernorZama.ts
@@ -7,8 +7,6 @@ import { createTransaction, produceDummyTransactions, waitForBlock } from '../ut
 import { deployCompFixture } from './Comp.fixture';
 import { deployGovernorZamaFixture, deployTimelockFixture } from './GovernorZama.fixture';
 
-const timeout = 180000;
-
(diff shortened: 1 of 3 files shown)


immortal-tofu:
You can define default timeout in `hardhat.config.ts`. Maybe it's better?

test/governor/GovernorZama.ts
```diff
@@ -35,7 +33,7 @@ describe('GovernorZama', function () {
 
   beforeEach(async function () {
     // Increase timeout for beforeEach
-    this.timeout(timeout);
+    this.timeout(180000);
```
dartdart26: Do we need to set to 180000 here?

dartdart26 (approved):

test/governor/GovernorZama.ts
david-zk: Not sure, will try rerunning tests without

david-zk: It passes, we can merge 👍

PR #221, opened 2023-11-08
feat() Implement feature explained in the article

immortal-tofu:


tremblaythibaultl (approved):
lgtm 👍

examples/Identity/AbstractCompliantERC20.sol
```diff
@@ -0,0 +1,59 @@
...
+    // Transfers an encrypted amount.
+    function _transfer(address from, address to, euint32 _amount) internal {
+        // Condition 1: hasEnoughFunds
+        ebool enoughFund = TFHE.le(_amount, balances[from]);
+        euint32 amount = TFHE.cmux(enoughFund, _amount, TFHE.asEuint32(0));
+
+        // Delegate call
+        (bool success, bytes memory returndata) = address(rulesContract).delegatecall(
+            abi.encodeWithSelector(ERC20Rules.transfer.selector, identityContract, rulesContract, from, to, amount)
+        );
```
dartdart26: I mentioned it offline, putting it here for reference. Why do we need `delegatecall` here? AFAICT, `ERC20Rules.transfer` doesn’t read storage from `AbtractCompliantERC20`. Or I miss something? Is it related to `msg.sender` and `tx.origin`?

examples/Identity/IdentityRegistry.sol
```diff
@@ -0,0 +1,140 @@
...
+
+    // A mapping from wallet to an identity.
+    mapping(address => Identity) internal identities;
+
+    struct Identity {
+        uint registrarId;
+        mapping(string => euint32) identifiers;
+    }
+
+    mapping(address => mapping(address => mapping(string => bool))) permissions; // users => contracts => identifiers[]
```
dartdart26, line 25: Do we need to have the final `mapping` or it could be an array instead as the comment suggests?

dartdart26 (approved):
Looks very good to me!

The only thing I didn't fully understand is the need for `delegatecall`.

examples/Identity/AbstractCompliantERC20.sol
immortal-tofu: I fixed ! :)

examples/Identity/IdentityRegistry.sol
immortal-tofu: No we need a mapping :(

PR #229, opened 2023-11-27
Fix ebool not + tests for ebool ops

david-zk:
Fixes ebool not, adds tests just in case also for and/or/xor

@immortal-tofu


leventdem (approved):
+1

PR #238, opened 2023-12-14
feat: add upper bound for encrypted random integers

dartdart26:
Add overloads of TFHE.randEuint() methods that accept an upper bound, e.g. TFHE.randEuint32(uint32 upperBound). The returned integer will be in the [0, upperBound) range. Moreover, upperBound must be a power of 2 - that is a design choice for performance reasons.

Note that, as of now, the random numbers are generated in the plain and are completely public and predictable. An FHE version is coming soon.

Add tests to cover above functionality.

### DIFF
--- a/test/rand/Rand.ts
+++ b/test/rand/Rand.ts
@@ -32,6 +32,20 @@ describe('Rand', function () {
     expect(unique.size).to.be.greaterThanOrEqual(2);
   });
 
+  it('8 bits generate with upper bound and decrypt', async function () {
+    const values: bigint[] = [];
+    for (let i = 0; i < 5; i++) {
+      const txn = await this.rand.generate8UpperBound(128);
+      await txn.wait();
+      const value = await this.rand.decrypt8();
+      expect(value).to.be.lessThanOrEqual(127);
+      values.push(value);
+    }
+    // Expect at least two different generated values.
+    const unique = new Set(values);
+    expect(unique.size).to.be.greaterThanOrEqual(2);
+  });
+
   it('16 bits generate and decrypt', async function () {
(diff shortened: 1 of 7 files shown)


dartdart26:
As of now, this PR needs the corresponding precompile in the backend and will fail CI tests.

test/rand/Rand.ts
```diff
@@ -72,42 +100,89 @@ describe('Rand', function () {
...
+
   it('16 bits in view', async function () {
     if (process.env.HARDHAT_NETWORK !== 'hardhat') {
       await expect(this.rand.generate16InView()).to.be.rejected;
     }
   });
 
+  it('16 bits with upper bound in view', async function () {
+    await expect(this.rand.generate16UpperBoundInView(128)).to.be.rejected;
+  });
```
jatZama, line 181: please add "if (process.env.HARDHAT_NETWORK !== 'hardhat')" so tests would also pass in Mocked version (when doing `pnpm test:mock`) in all tests related to "generateXUpperBoundInView"

jatZama: Also, the Mocks/Impl.sol file must be updated with a mocked `randBounded` function btw

immortal-tofu: There was an issue with the path of mocks, rebase before adding mock to avoid issue :)

dartdart26: Done.

jat9292 (approved):
lgtm

test/rand/Rand.ts
jatZama: lgtm now

jatZama (approved):

PR #239, opened 2023-12-20
made changes for correcting shift operators in tfhe sol

jatZama:
there is still some error that I do not understand where it comes from in the case of shift operators with : (uintX,euint8) signature. Codegen works but when running tfheOperations test with hardhat it tells me : 
```
TypeError: Member "shl" not unique after argument-dependent lookup in type(library TFHE).
  --> examples/tests/TFHETestSuite3.sol:10:25:
   |
10 |         euint8 result = TFHE.shl(aProc, bProc);
   |                         ^^^^^^^^
```

Which does not make sens for me since this function is defined only once with the (uint8,euint8) signature inside TFHE.sol

EDIT : this was solved by removing the shl/shr operators with (uintX,euitn8) signatures. Because solidity does not support this type of overloading :  https://forum.soliditylang.org/t/no-unique-declaration-found-when-overloading-functions-with-integers-of-different-sizes/1308 

In conclusion this PR removes the shift operators with a second operand greater than euint8/uint8 types.
By adding a `shiftOperator` key in the two shift operators objects, it implements the following operators : 
`shr(euintX,euint8)`, `shl(euintX,euint8)`, `shr(euintX,uint8)` and `shl(euintX,uint8)`, whenre `X` is `8`, `16` or `32`.

### DIFF
--- a/codegen/templates.ts
+++ b/codegen/templates.ts
@@ -473,7 +473,7 @@ function tfheAsEboolUnaryCast(bits: number): string {
     }
 
     function not(ebool a) internal pure returns (ebool) {
-        return asEbool(not(asEuint8(a)));
+        return asEbool(and(not(asEuint8(a)), asEuint8(1)));
     }
     
     // If 'control''s value is 'true', the result has the same value as 'a'.
--- a/codegen/templates.ts
+++ b/codegen/templates.ts
@@ -280,8 +280,8 @@ function tfheEncryptedOperator(
     operator.returnType == ReturnType.Uint
       ? `euint${outputBits}`
       : operator.returnType == ReturnType.Ebool
-        ? `ebool`
-        : assert(false, 'Unknown return type');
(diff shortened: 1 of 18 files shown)


aquint-zama:
@cla-bot check

immortal-tofu:
You should clean your PR. It seems you have integrated previously merged commit concerning package.json and docs.

codegen/templates.ts
```diff
@@ -473,7 +556,7 @@ function tfheAsEboolUnaryCast(bits: number): string {
     }
 
     function not(ebool a) internal pure returns (ebool) {
-        return asEbool(not(asEuint8(a)));
+        return asEbool(and(not(asEuint8(a)), asEuint8(1)));
```
david-zk, line 559: Hmmm this change is from previous commit, did you branch from the current master when making this PR?

PR #242, opened 2023-12-27
Mocked version of TFHE.sol for mocked tests and coverage

jatZama:
This branch add a mocked version of TFHE.sol and Impl.sol
All tests should pass, as is, passing without any modification, neither the javascript files neither the solidity files need to be changed. Actually, very minor modifications were needed in few .js files (but 0 modification on the solidity side). The codegen also produces the needed mocked solidity files for the mocked library.
To run the mocked tests : `pnpm test:mock` - all tests should pass in 3 seconds in this case instead of the original 21 minutes, allowing a better developer experience.
To get the coverage of tests in mocked mode : `pnpm coverage:mock`, then open the file `coverage/index.html`. This will allow increased security by pointing out missing branches not covered yet by the current test suite.

Notice that, due to intrinsic limitations of the original EVM, the mocked version differ in few corner cases from the real fhevm, the most important change is the `TFHE.isInitialized` method which will always return true in the mocked version. This means that before deploying to production, developpers would still need to run the tests with the original fhevm node, as a final check in non-mocked mode, with `pnpm test`.

### DIFF
--- a/codegen/main.ts
+++ b/codegen/main.ts
@@ -9,10 +9,13 @@ function generateAllFiles() {
 
   const network = Network[(process.env.TARGET_NETWORK as keyof typeof Network) || 'Evmos'];
   const context = networkCodegenContext(network);
-  const [tfheSolSource, overloads] = t.tfheSol(context, operators, SUPPORTED_BITS);
+  const [tfheSolSource, overloads] = t.tfheSol(context, operators, SUPPORTED_BITS, false);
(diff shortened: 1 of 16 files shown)


immortal-tofu:
Great work! Only suggestion would be to rename `lib_mock` to `mocks` because we'll publish the mock.
The next step will be to update the hardhat template with the preprocessor part but we need to publish the library and mock before.

immortal-tofu (approved):

PR #244, opened 2024-01-05
Updated Solidity version to ^0.8.20 and OpenZeppelin Contracts to ^5.0.1

jatZama:


jatZama:
@leventdem There is something fishy here, CI is still running after 2H30 already on the tests and I am unable to see the logs. Using ethermint-dev-node:v0.2.1
Edit : I killed the CI tests because I figured out version 0.2.1 was not released yet. Weird that I did not get an error.

docs/howto/write_contract/hardhat.md
```diff
@@ -1,7 +1,5 @@
 # Using Hardhat
 
-> Our library TFHE requires Solidity version **0.8.19** specifically, as we rely on features exclusive to this version and do not currently provide support for versions beyond it.
```
immortal-tofu, line 3: Good catch 👍

immortal-tofu (approved):
LGTM when CI will be ok

test/utils.ts
```diff
@@ -5,41 +5,49 @@ import type { Counter } from '../types';
...
-          resolve(currentBlock);
+          reject(error);
         }
-      } catch (error) {
-        clearInterval(intervalId);
-        reject(error);
-      }
-    }, 50); // Check every 50 milliseconds
-  });
+      }, 50); // Check every 50 milliseconds
```
immortal-tofu, line 22: Why don't you use the same script as below with the block event? It seems better, performance-wise, to trigger evaluation on block event instead of spamming the function every 50ms?

jatZama: This is only when using mocked mode, i.e as you can see when HARDHAT_NETWORK env variable is set to "hardhat", in this case we are using hardhat node so there is no mining each 15 seconds, but only automining when there is a new transaction or a manual trigger via the `mineNBlocks` function from `utils.ts`. This part was already implemented when I merged the mocked version fyi. It is necessary to get the governor tests passing in mocked mode btw.

PR #245, opened 2024-01-05
Feature/improved erc20

jatZama:
For the moment this is a draft PR, which should be completed early next week, needing first this PR to be accepted : PR #244

First some general remarks. The EncryptedERC20 cannot implement the full [EIP20 specifications](https://eips.ethereum.org/EIPS/eip-20). For example a function with signature balanceOf(address _owner) does not make sense, since only the msg.sender should be able to decrypt his own balance, so we need to add at least a bytes32 and a bytes argument to use the onlySignedPublicKey modifier correctly. Similarly, because amounts are encrypted, an event with the following signature does not make sense : event Transfer(address indexed _from, address indexed _to, uint256 _value). The value_ argument is meaningless in the case of an EncryptedEC20. This is why, some changes compared to the original standard need to be discussed first before implementation (see “Possible further improvements” section later below).

Let’s start with the most important remark : I noticed also a critical bug which needs to me corrected urgently in the transferFrom function: there is currently no real check on the allowance of a spender, so anyone could spend anyone’s else tokens!! 
This is a POC showing the vulnerability : 
```js
  it('should mint the contract', async function () {
    const encryptedAmount = this.instances.alice.encrypt32(1000);
    const transaction = await this.erc20.mint(encryptedAmount);
    await transaction.wait();
    // Call the method
    const token = this.instances.alice.getTokenSignature(this.contractAddress) || {
      signature: '',
      publicKey: '',
    };
    const encryptedBalance = await this.erc20.balanceOf(this.signers.alice, token.publicKey, token.signature);
    // Decrypt the balance
    const balance = this.instances.alice.decrypt(this.contractAddress, encryptedBalance);
    expect(balance).to.equal(1000);

    const encryptedTotalSupply = await this.erc20.getTotalSupply(token.publicKey, token.signature);
    // Decrypt the total supply
    const totalSupply = this.instances.alice.decrypt(this.contractAddress, encryptedTotalSupply);
    expect(totalSupply).to.equal(1000);

    console.log("--- BUG : BOB can transferFrom Alice's funds without approval! ---");
    const tx = await this.erc20.connect(this.signers.bob)['transferFrom(address,address,bytes)'](this.signers.alice.address, this.signers.bob.address, encryptedAmount);
    await tx.wait();
    const encryptedBalance2 = await this.erc20.balanceOf(this.signers.alice, token.publicKey, token.signature);
    // Decrypt the balance
    const balance2 = this.instances.alice.decrypt(this.contractAddress, encryptedBalance2);
    expect(balance2).to.equal(0);
  });

```

What this PR contains already : 
1/ Made totalSupply public and unencrpyted, since this value was never confidential to begin with (anyone could query it via getTotalSupply).
2/ Removed getTotalSupply (because of previous point)
3/ Modified the mint function to take now an unencrypted uint32 argument. This also has the benefits of solving a previously possible overflow error when minting new tokens, which could have been very dangerous because could have gone unnoticed. 
4/ Changed decimals to 0
5/ removed unused DOMAIN_SEPARATOR
6/ Made contractOwner, NO_ERROR and NOT_ENOUGH_FUND immutable.
7/ Simplified balanceOf()and renamed it to myBalance()
8/ Improved the allowance() function by making it callable by either the owner or the spender, instead of just the owner originally, which is more sensible and makes it more useful this way.

Possible further improvements needing discussion before implementation: 
1/ Fix the critical bug on transferFrom/ . TODO on Monday : this can be fixed easily for instance by passing an additional encrypted boolean variable to the _transfer function, depending on the result of the approval. Notice : the _updateAllowance() function also needs fixing, especially the _approve call at the end.
2/ All tests must be modified accordingly to pass with this new version.
3/ Add simplified Transfer, Approval and Mint events, without the amounts (because they are encrypted), for better interactivity offchain and to follow more closely the spirit of the original EIP20.
4/ Make the transfer, transferFrom and approve functions return booleans, to make the EncryptedERC20 contract adhere more closely to the original EIP20 standard.
5/ Instead of a unique owner which could not renounce his ownership, inherit from Ownable2Step from OZ library.
6/ I am not a fan of the current error handling system. For example, if a smart contract account instead of an EOA is used, several transfers could happen in a single tx. But with current implementation of getLastError(), only the very last error would become available for decryption. There is also the fact that currently we are unable to detect if a transfer has failed because of an insufficient allowance. There is a lot of room to improve on this.
7/ Obviously many extensions are possible : for example, a wrapper contract to convert ETH/native token or any regular unencrypted ERC20 to an encrypted version, adding a burn function, adding a callback function like in the
ERC-777 standard, some even argue that this should be the default standard for ERC20 because it allows better UX (single tx for swapping instead of approve + transfer) despite risk of reentrancy.


examples/EncryptedERC20.sol
```diff
@@ -43,10 +40,10 @@ contract EncryptedERC20 is Reencrypt {
     }
 
     // Sets the balance of the owner to the given encrypted balance.
-    function mint(bytes calldata encryptedAmount) public onlyContractOwner {
-        euint32 amount = TFHE.asEuint32(encryptedAmount);
+    function mint(uint32 mintedAmount) public onlyContractOwner {
```
immortal-tofu: The mint need to be encrypted. But there is an open question how we can obfuscate the mint :)

examples/EncryptedERC20.sol
```diff
@@ -43,10 +40,10 @@ contract EncryptedERC20 is Reencrypt {
...
+        euint32 amount = TFHE.asEuint32(mintedAmount);
```
immortal-tofu: You don't need to transform a uint32 into a euint32 to add it. It will be faster to do a `TFHE.add(balances[contractOwner], mintedAmount)` than a `TFHE.add(balances[contractOwner], TFHE.asEuint32(mintedAmount))`

examples/EncryptedERC20.sol
```diff
@@ -59,24 +56,13 @@ contract EncryptedERC20 is Reencrypt {
         _transfer(msg.sender, to, amount);
     }
 
-    function getTotalSupply(
```
immortal-tofu, line 62: Needed if we want to get encrypted mint

examples/EncryptedERC20.sol
```diff
@@ -59,24 +56,13 @@ contract EncryptedERC20 is Reencrypt {
...
-        bytes calldata signature
-    ) public view onlySignedPublicKey(publicKey, signature) returns (bytes memory) {
-        return TFHE.reencrypt(totalSupply, publicKey, 0);
-    }
-
     // Returns the balance of the caller encrypted under the provided public key.
 
-    function balanceOf(
-        address wallet,
+    function myBalance(
```
immortal-tofu: I never saw `myBalance`? ERC20 uses balanceOf, no? https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol

immortal-tofu (commented):
Few adjustments. The PR mix update of Solidity version and ERC20 and need to be clean to see what would be really merged.

immortal-tofu:
Regarding mint, encrypted mint is useless today because you can get totalSupply before and after. We discuss about the idea of giving a range for totalSupply instead of exact value, but we didn't found a real solution.
In the meantime, I think we can adopt what you did and let mint amount and total supply in plaintext.

jatZama:
> Few adjustments. The PR mix update of Solidity version and ERC20 and need to be clean to see what would be really merged.

As I said in the first sentence this draft PR will come after "needing first this PR to be accepted : PR https://github.com/zama-ai/fhevm/pull/244" which updates the OZ and Solidity versions (was blocked yesterday because of the docker image not being public then the PUSH0 problem, but should be merged asap)

examples/EncryptedERC20.sol
jatZama: This is following my first paragraph fyi : "The EncryptedERC20 cannot implement the full [EIP20 specifications](https://eips.ethereum.org/EIPS/eip-20). For example a function with signature balanceOf(address _owner) does not make sense, since only the msg.sender should be able to decrypt his own balance, so we need to add at least a bytes32 and a bytes argument to use the onlySignedPublicKey modifier correctly." But I am open to discussion if you think a better name is acceptable

examples/EncryptedERC20.sol
jatZama: This is following this the remark point number 1/ from my PR description : "1/ Made totalSupply public and unencrpyted, since this value was never confidential to begin with (anyone could query it via getTotalSupply)." 
I think you must chose one of 2 options here, you cannot have both : 
1/ Either keep mints encrypted, in this case getTotalSupply could not exist (at least as is, maybe it is possible to have a partial version if we rate limit its calls, to get differential/partial privacy). In this case it would be better to add an overflow check when minting.
2/ If you want to keep getTotalSupply then because of my remark in point number 1/, there is no need to keep the mints encrypted as this would leak the confidentiality of the mints.

examples/EncryptedERC20.sol
jatZama: Good point.

examples/EncryptedERC20.sol
immortal-tofu: I see your point but @randhindi wanted to keep balanceOf signature to stick to EIP20 specifications and avoid the introduction of a new method name. To be honest, I don't have strong opinion on this.

jatZama: We can keep a dummy balanceOf function which would return the euint32 value handle in this case, but also add a myBalance function with the new signature!

immortal-tofu: Hmm no because if you return the euint32 without checking any signature, you allow anyone to decrypt the value

jatZama: Ah yes very good point, maybe returning a dummy value in the case where msg.sender != owner argument would be enough...

examples/EncryptedERC20.sol
immortal-tofu: Let's switch to unencrypted mint and make type totalSupply with uint32 👍

examples/EncryptedERC20.sol
immortal-tofu: I think we should keep the "balanceOf(address)" function, as we may need it in the future for a "viewer" delegation feature similar to the existing "approve" functionality for spend.
It would also make a lot of sense because we'll have authorization decryption associated to a wallet in a near future.

jatZama: yes my point was indeed to keep it (while also adding a myBalance() function) . 
`balanceOf(address)` would work similarly to `myBalance()` if and only if address==msg.sender , otherwise it would return 0 for instance. Wdyt?

jatZama:
EDIT : Following discussion with @immortal-tofu on Slack, this PR has been updated with the following updates, in addition to the first post : 
1/ Major fix of transferFrom / allowance bug.
2/ Removed all error handling form EncryptedERC20 for the moment.
3/ Tests have been updated accordingly (from EncryptedERC20 and Blind Auction tests). Also added a test to check correctness of `transferFrom` function. 
4/ Added a second version of the `approve` public function taking a `euint32`as second argument.
5/ Added Ownable2Step import (hence, removed the `contractOwner` state variable)
6/ Minted amounts and totalSupply are now in cleartext.
7/ Added boolean returns for `approve` , `transfer` and `transferFrom` functions to follow more closely the EIP 20 standard.
8/ Added `Transfer`, `Approval` and `Mint` events.
9/ Kept previous version of `balanceOf` and removed he proposed `myBalance` function.

jatZama:
Just awaiting a review before merging @immortal-tofu @leventdem @david-zama @dartdart26 🙏

jatZama:
After further discussion with Clément, I also rewrote the `CompliantERC20` which had similar code and issues to previous version of `EncryptedERC20`.  This is the second commit of this PR. In summary it contains : 
1/ Removal of `AbstractCompliantERC20` abstract contract.
2/ `CompliantERC20` now inherits from `EncrpytedERC20`, with added `virtual` modifiers for `balanceOf` and `_transfer` functions because they are overriden inside `CompliantERC20`.
3/ Fixed a logic error inside `ERC20Rules` in the `checkBlacklist` function.
4/ Fixed corresponding tests of `CompliantERC20`.

examples/EncryptedERC20.sol
```diff
@@ -3,110 +3,98 @@
...
-        NOT_ENOUGH_FUND = TFHE.asEuint8(1);
-    }
-
-    function getLastError(
-        bytes32 publicKey,
-        bytes calldata signature
-    ) public view onlySignedPublicKey(publicKey, signature) returns (bytes memory) {
-        return TFHE.reencrypt(lastError[msg.sender], publicKey, 0);
-    }
+    constructor() Ownable(msg.sender) {}
```
immortal-tofu, line 25: Ownable() or Ownable2Step()?

immortal-tofu (approved):
Small comment otherwise LGTM

examples/EncryptedERC20.sol
jatZama: I also thought that `Ownable2Step` had to be placed in the constructor at first, but Solidity did not compile in this case... in fact, `Ownable2Step` also inherits from `Ownable`, and it is the constructor of `Ownable` that originally contains the initialOwner argument, so I think that's how Solidity works.

jatZama: Actually figured out this due to the fact that `Ownable2Step` is not really a contract but just an abstract contract https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable2Step.sol as you can see it inherits from `Ownable` but does not contain a constructor! 
This is why in the constructor of `EncryptedERC20` you still need to use `Ownable(msg.sender)` and not `Ownable2Step(msg.sender)` since the latter does not exist actually !

jatZama: Made some quick tests in Remix to illustrate this : 
This compiles : 
Screenshot 2024-01-10 at 15 51 11
But this does not compile : 
Screenshot 2024-01-10 at 15 52 09

On the other hand, if Own2S was a real constract (not abstract) with its own constructor, then `constructor(address initOwner_) Own2S(initOwner_)` would have to be used in the Token contract in order to make it compile. Hope this is clear for you. Eg this compiles : 
Screenshot 2024-01-10 at 15 56 11

PR #249, opened 2024-01-15
Added NonceManager to avoid Nonce errors in CI tests

jatZama:
I wrapped the signers inside NonceManager objects, following this issue : https://github.com/ethers-io/ethers.js/issues/4258#issuecomment-1705709167 
This should solve hopefully the occasional bug in CI due to Nonce reuse when jumping to new test files when running the test suite.


jatZama:
CI tests successfully passed twice in a row fyi. Awaiting your review @immortal-tofu please

immortal-tofu (approved):

PR #250, opened 2024-01-15
fix() update gas prices in doc

immortal-tofu:


docs/howto/gas.md
```diff
@@ -2,31 +2,36 @@
 
 FHE operations are typically more computationally expensive than classical operations due to their inherent complexity. As a reference, here is an approximation of the gas cost associated with each operation.
```
jatZama, line 3: Are these values only "approximation"s or the exact ones chosen for the moment?

docs/howto/gas.md
```diff
@@ -31,7 +31,7 @@ FHE operations are typically more computationally expensive than classical opera
 | `TFHE.not`          | 29,000       | 31,000  | 33,000  |
 | `TFHE.cmux`         | 60,000       | 65,000  | 70,000  |
 | `TFHE.decrypt()`    | 500,000      | 500,000 | 500,000 |
-| `TFHE.randEuintX()` | 100,000      | 200,000 | 400,000 |
+| `TFHE.randEuintX()` | 100,000      | 100,000 | 100,000 |
```
jatZama, line 34: | `TFHE.randEuintX()` | 100,000      | 100,000 | 100,000 | --> seems conter-intuitive for me , not sure how is this possible?

immortal-tofu: Randomness of 1 bit is 1 pbs. Since it's parallelized, the timing is the same and pretty fast. Rand expects to be as fast as one `and`. For now, we keep room for improvement

docs/howto/gas.md
immortal-tofu: The exact chosen for the moment

jat9292 (approved):

PR #253, opened 2024-01-16
docs() add troubleshooting regarding property

immortal-tofu:


docs/howto/types.md
jatZama, line 1: Instead of "property" I would use "state variable" as this is the most common used terminology.
Also could you use a warning such as : "Aslo please be aware that you should never declare encrypted types as neither `constant` or `immutable` types, as this variables need to actually stored in the privileiged storage of the fhEVM which is not compatible with constant variable types".

jatZama (changes requested):
Instead of "property" I would use "state variable" as this is the most commonly used terminology.
Also could you use a warning such as : "Aslo please be aware that you should never declare encrypted types as neither constant or immutable types, as this variables need to actually stored in the privileiged storage of the fhEVM which is not compatible with constant variable types".

docs/howto/types.md
```diff
@@ -51,7 +51,8 @@ function mint(bytes calldata encryptedAmount) public onlyContractOwner {
 
 ## Contract property with encrypted types
```
jatZama: ## Contract state variables with encrypted types

docs/howto/types.md
```diff
@@ -48,3 +48,16 @@ function mint(bytes calldata encryptedAmount) public onlyContractOwner {
   totalSupply = totalSupply + amount;
 }
 ```
+
+## Contract property with encrypted types
+
+If you require a state variable that utilizes these encrypted types, you cannot directly assign the value. In Solidity, the compiler attempts to ascertain the value of `TFHE.asEuintXX(yy)` during compilation, which is not feasible because `asEuintXX()` invokes a precompiled contract. To address this challenge, you should declare your property and subsequently assign its value within the constructor.
+Also, please be aware that you should never declare encrypted types as either constant or immutable types, as these variables need to be stored in the privileged storage of the fhEVM, which is not compatible with constant variable types.
```
jatZama: "Also, please be aware that you should never declare encrypted types as either constant or immutable variables"

jatZama (approved):

PR #261, opened 2024-01-18
Release/0.3.0

immortal-tofu:


tasks/utils.ts
```diff
@@ -25,21 +25,21 @@ export const createInstance = async (contractAddress: string, account: Signer, e
...
     verifyingContract: contractAddress,
   });
 
   // Sign the public key
   const signature = await signer.signTypedData(
-    generatedToken.token.domain,
-    { Reencrypt: generatedToken.token.types.Reencrypt }, // Need to remove EIP712Domain from types
-    generatedToken.token.message,
+    generatedToken.eip712.domain,
+    { Reencrypt: generatedToken.eip712.types.Reencrypt }, // Need to remove EIP712Domain from types
```
jatZama, line 41: small remark : I dont understand the meaning of the comment " Need to remove EIP712Domain from types" which confuses me a little

jatZama (approved):

tasks/utils.ts
immortal-tofu: In the EIP712 standard, the `types` contains type descriptions of your object. With Metamask, it must contain the type description of `EIP712Domain` key you are using (the `domain` key in the object). On the other hand, if you sign with ethers, ethers don't recognize this `EIP712Domain` type and throw an error. You need to remove it, probably because it doesn't expect a type definition for `domain` (because this is in the EIP712 standard)

PR #276, opened 2024-02-01
Added documentation for pitfalls and best practises

jatZama:


docs/howto/pitfalls.md
```diff
@@ -0,0 +1,258 @@
...
+    b = TFHE.asEuint32(_b);
+  }
+}
+```
+
+✅ Should be replaced by this snippet:
+
+```solidity
+contract C {
+  euint32 internal a = TFHE.asEuint32(42);
```
immortal-tofu, line 24: No, you can't call precompiles on properties

docs/howto/pitfalls.md
```diff
@@ -0,0 +1,258 @@
...
+
+  constructor(uint32 _b) {
+    b = TFHE.asEuint32(_b);
+  }
+}
+```
+
+## Never use public encrypted state variables
+
+Or else, this would expose the variabls to any external untrusted smart contract which could ask for decrypting them, losing confidentiality.
```
immortal-tofu: `Declaring an encrypted state variable as public exposes the variable to any external untrusted smart contract to access and potentially decrypt them, compromising their confidentiality.`

docs/howto/pitfalls.md
```diff
@@ -0,0 +1,258 @@
...
+    a = TFHE.asEuint32(_a);
+  }
+}
+```
+
+In this last snippet, the `internal` keyword could have been omitted (satate variables are private by default) or alternatively have been replaced by `private`.
+
+## Protect access of view functions using reencryptions
+
+If a view function is using `TFHE.reencrypt` it is mandatory to protect it via the `onlySignedPublicKey` modifier imported from `"fhevm/abstracts/Reencrypt.sol"`. See the example from the [Decrypt page](decrypt.md#handle-private-reencryption). If you fail to do this, then anyone could reencrpyt anyone else's ciphertext, because when doing a static call to a view function you could impersonate any `msg.sender`'s address since no signature is needed, contrarily to transactions.
```
immortal-tofu: `If you fail to do this, then anyone could reencrpyt anyone else's ciphertext, because when doing a static call to a view function you could impersonate any `msg.sender`'s address since no signature is needed, contrarily to transactions.`
=>
`Failing to address this allows anyone to reencrypt another person's ciphertext. This vulnerability comes from the ability to impersonate any msg.sender address during a static call to a view function, as it does not require a signature, unlike transactions.`

docs/howto/pitfalls.md
```diff
@@ -0,0 +1,258 @@
...
+
+# Best practises
+
+## Avoid using TFHE.decrypt, use CMUX instead when possible
```
immortal-tofu: `Avoid using TFHE.decrypt, use TFHE.cmux instead`

docs/howto/pitfalls.md
```diff
@@ -0,0 +1,258 @@
...
+
+Any use of decryption should be avoided as much as possible. Current version of `TFHE.decrypt` will soon be deprecated and get replaced by an asynchronous version, so please consider this operator as a very expensive one which should be used only if absolutely necessary.
+
+Please also not that the optimistic require operator `TFHE.optReq` is also deprecated: we recommend to avoid its use in all cases.
```
immortal-tofu: I'd remove this sentence because this is not related to the topic (decrypt).

Create a specific section for optReq or simply remove all mentions from the documentation?

docs/howto/pitfalls.md
```diff
@@ -0,0 +1,258 @@
...
+
+```solidity
+euint32 x;
+ebool condition = TFHE.gt(x,5)
+x = TFHE.cmux(condition, TFHE.asEuint(0), TFHE.asEuint(42));
+```
+
+## Obfuscate branching
+
+In the previous paragraph, we already saw that branching logic depending on ciphertexts should rely as much as possible on CMUX operators instead of decryptions, in order to hide which branch has been executed effectively by the transaction.
```
immortal-tofu: `The previous paragraph emphasized that branch logic should rely as much as possible on TFHE.cmux instead of decryptions. It hides effectively which branch has been executed.`

docs/howto/pitfalls.md
```diff
@@ -0,0 +1,258 @@
...
+
+But this is sometimes not enough. Naive application of FHE operations can still leak a lot of information. Adding privacy to smart contracts require revisiting sometimes the logic of your application.
```
immortal-tofu: `However, this is sometimes not enough. Enhancing the privacy of smart contracts often requires revisiting your application's logic.`

docs/howto/pitfalls.md
```diff
@@ -0,0 +1,258 @@
...
+ebool isTrue;
+euint32 x;
+// some code
+while(TFHE.decrypt(isTrue)){
+    x=TFHE.add(x, 1);
+    // some other code
+}
+```
+
+If your code logic requires looping on an encrypted boolean condition, we highly suggest to try to replace it by a finite loop with an appropriate constant maximum number of steps and use CMUX inside the loop.
```
immortal-tofu: `TFHE.cmux`

docs/howto/pitfalls.md
```diff
@@ -0,0 +1,258 @@
...
+    x=TFHE.add(x, increment);
+    // some other code
+}
+```
+
+## Avoid using encrypted indexes
+
+Using encrypted indexes to pick an element from an array without revealing it is not very efficient, because you would still need to loop on all the indexes to preserve confidentiality.
+
+However, there are plans to make this kind of operation much more efficient on the roadmap, using a specialized operator for arrays.
```
immortal-tofu: I wouldn't talk about roadmap here because we'll probably forget to update this here. We have a specific roadmap page for this.

docs/howto/pitfalls.md
```diff
@@ -0,0 +1,258 @@
...
+
+For example, let's say you have an encrypted array variable `encArray` and want to set some encrypted state variable `x` to be equal to some element `encArray[i]`, _without_ disclosing the index `i`.
```
immortal-tofu: `For instance, imagine you have an encrypted array called `encArray` and you want to update an encrypted value `x` to match an item from this list, encArray[i], without disclosing which item you're choosing.`

docs/howto/pitfalls.md
```diff
@@ -0,0 +1,258 @@
...
+
+❌ For the moment, you must loop over all the indexes and check equality homomorphically, however this pattern is very expensive in gas and should be avoided whenever possible.
```
immortal-tofu: Remove `For the moment`

`You must loop over all the indexes and check equality homomorphically, however this pattern is very expensive in gas and should be avoided whenever possible.`

docs/howto/pitfalls.md
```diff
@@ -0,0 +1,258 @@
...
+
+```solidity
+function mint(bytes calldata encryptedAmount) public {
+  euint32 mintedAmount = TFHE.asEuint32(encryptedAmount);
+  totalSupply = TFHE.add(totalSupply, mintedAmount);
+  balances[msg.sender] = TFHE.add(balances[msg.sender], mintedAmount);
+}
+```
+
+✅ But you can fix this issue by using CMUX to cancel the mint in case of an overflow:
```
immortal-tofu: `TFHE.cmux`

docs/howto/pitfalls.md
```diff
@@ -0,0 +1,258 @@
...
+  ebool isOverflow = TFHE.lt(tempTotalSupply, totalSupply);
+  totalSupply = TFHE.cmux(isOverflow, totalSupply, tempTotalSupply);
+  euint32 tempBalanceOf = TFHE.add(balances[msg.sender], mintedAmount);
+  balances[msg.sender] = TFHE.cmux(isOverflow, balances[msg.sender], tempBalanceOf);
+}
+```
+
+Notice that we did not check separately the overflow on `balances[msg.sender]` but only on `totalSupply` variable, because `totalSupply` is the sum of the balances of all the users, so `balances[msg.sender]` could never overflow if `totalSupply` did not.
+
+# Typical workflow for writing confidential smart contracts on the fhEVM
```
immortal-tofu: It's probably something we should rewrite and move to the `Getting started` page. Something which link to the subpages hardhat/remix/...

docs/howto/pitfalls.md
```diff
@@ -0,0 +1,258 @@
...
+
+```
+
+Notice that to preserve confidentiality, we had to make two inputs transfers on both tokens from the user to the AMM contract, and similarly two output transfers from the AMM to the user, even if technically most of the times it will make sense that one of the user inputs `encryptedAmountAIn` or `encryptedAmountBIn` is actually an encrypted zero.
+
+This is different from a classical non-confidential AMM with regular ERC20 tokens: in this case, the user would need to just do one input transfer to the AMM on the token being sold, and receive only one output transfer from the AMM on the token being bought.
+
+## Avoid using while loops with an encrypted condition
+
+❌ This kind of loop should be absolutely avoided, because it relies potentially on a huge number of decrypts, and even a single decrypt should be avoided in general:
```
immortal-tofu: `Avoid using this type of loop because it might require many decryption operations:`

docs/howto/pitfalls.md
```diff
@@ -0,0 +1,258 @@
...
+contract C {
+  euint32 public a;
+
+  constructor(uint32 _a) {
+    a = TFHE.asEuint32(_a);
+  }
+}
+```
+
+✅ But write instead:
```
immortal-tofu: Instead, you should describe your property as follow:

docs/howto/pitfalls.md
```diff
@@ -0,0 +1,258 @@
...
+        x = TFHE.cmux(isEqual, encArray[i], x);
+    }
+}
+```
+
+## Use scalar operands when possible to save gas
+
+Some TFHE operators exist in two versions : one where all operands are ciphertexts handles, and another where one of the operands is an unencrypted scalar. Whenever possible, use the scalar operand version, as this will save a lot of gas. See the page on [Gas](gas.md) to discover which operators support scalar operands and compare the gas saved between both versions: all-encrypted operands vs scalar.
+
+This means that you should avoid trivial encryptions when possible.
```
immortal-tofu: I'd remove this sentence, seems redudant

immortal-tofu (changes requested):

docs/howto/pitfalls.md
jatZama: I do not understand your review : I just tried this code and it works without problem on Zama Remix IDE fyi. 
```solidity
pragma solidity ^0.8.20;

import "fhevm/lib/TFHE.sol";

contract C {
    euint32 internal a = TFHE.asEuint32(42);
    function getA() public view returns(uint32) {
        return TFHE.decrypt(a);
    }
}
```

The important thing to make everything working is to remove the constant/immutable keyword to let the TFHE computation and storage in privileged storage happen at construction time.

immortal-tofu: Ho I didn't know it was just the keywords `immutable`/`constant` blocking. I thought it was totally forbidden. My bad in this case 🙏

docs/howto/pitfalls.md
jatZama: "Instead, you should declare the state variable as follow:" seems even better here, wdyt?

docs/getting_started/workflow.md
```diff
@@ -0,0 +1,13 @@
+# Typical workflow for writing confidential smart contracts
```
immortal-tofu: Sorry I wasn't clear: what I thought was to add this section in the getting started page for contract => https://docs.zama.ai/fhevm/how-to/write_contract
It could be a section below `Installation` where we introduce the 2 subpages (Hardhat and Remix) and emphasizes that developer should use the template. We have so many issues because users start with their own custom hardhat template...

If we need to elaborate about launching hardhat, we can add this in the [Hardhat template page](https://docs.zama.ai/fhevm/how-to/write_contract/hardhat). Like how to use mocks, how to start the docker image, etc etc... I wouldn't mix TFHE.cmux best practice in this section for example.

Finally, if we want to explain how to write a contract, we can write a full tutorial (writing without encryption and then write with Remix, then write in hardhat, then write test, ...). It could be a full video or an article.

docs/SUMMARY.md
```diff
@@ -10,6 +10,7 @@
 - [Connecting to Zama Devnet](getting_started/connect.md)
 - [Using Zama Faucet](getting_started/faucet.md)
 - [Local dev node](getting_started/local.md)
+- [Typical workflow for writing confidential smart contracts](getting_started/workflow.md)
```
immortal-tofu: To remove, see other comment

immortal-tofu (changes requested):

docs/howto/write_contract.md
```diff
@@ -20,3 +20,13 @@ pnpm add fhevm
...
+
+2/ A good first step is to start with an unencrypted version of the contract you want to implement, as you would usually do on a regular EVM chain. It is easier to reason first on cleartext variables, before thinking on how to add confidentialy. You could test your contracts on a regular hardhat node via : `npx hardhat test --network hardhat`.
+
+3/ Then launch an fhEVM local node with `pnpm fhevm:start` and start to add confidentiality by using the `TFHE` solidity library. Typically, this would involve converting some `uintX` types to `euintX`, as well as following all the detailed advices that we gave in the [pitfalls to avoid and best practises](../howto/pitfalls.md) section of the documentation. For inspiration, you can look at the examples inside the [`fhevm` repository](https://github.com/zama-ai/fhevm/tree/main/examples). Before using the hardhat template, please alse read the advices that we gave in the [hardhat section](./write_contract/hardhat.md).
```
immortal-tofu:
```
When you're ready, you can start to add confidentiality by using the `TFHE` solidity library. Typically, this would involve converting some `uintX` types to `euintX`, as well as following all the detailed advices that we gave in the [pitfalls to avoid and best practises](../howto/pitfalls.md) section of the documentation. For inspiration, you can take a look at the examples inside the [`fhevm` repository](https://github.com/zama-ai/fhevm/tree/main/examples). If you're using the Hardhat template, read the advices that we gave in the [Hardhat section](./write_contract/hardhat.md).
```

docs/howto/write_contract.md
```diff
@@ -20,3 +20,13 @@ pnpm add fhevm
 ```
 
 This will download and install the fhEVM Solidity Library and its dependencies into your project.
+
+## Typical workflow for writing confidential smart contracts
+
+We present here a quick sketch of a typical workflow for writing confidential smart contracts on the fhEVM.
+
+1/ For quick prototyping of a specific feature, use the [Zama version of the Remix IDE](./write_contract/remix.md). This will let you quickly deploy a contract on the devnet via Metamask, and interact easily with it throug the Remix UI. Otherwise, for a bigger project, you should start by cloning our custom [`fhevm-hardhat-template` repository](https://github.com/zama-ai/fhevm-hardhat-template). Hardhat is a popular development environment for Solidity developers and will let you test and deploy your contracts to the fhEVM using Javascript or Typescript code.
```
immortal-tofu:
```
For quick prototyping of a specific feature, use the [Zama version of the Remix IDE](./write_contract/remix.md). This will let you quickly deploy a contract on the devnet via Metamask, and interact easily with it through the Remix UI.
Otherwise, for a bigger project, you should use our custom [`fhevm-hardhat-template` repository](https://github.com/zama-ai/fhevm-hardhat-template). Hardhat is a popular development environment for Solidity developers and will let you test and deploy your contracts to the fhEVM using TypeScript.
```

docs/howto/write_contract.md
```diff
@@ -20,3 +20,13 @@ pnpm add fhevm
 ```
 
 This will download and install the fhEVM Solidity Library and its dependencies into your project.
+
+## Typical workflow for writing confidential smart contracts
+
+We present here a quick sketch of a typical workflow for writing confidential smart contracts on the fhEVM.
```
immortal-tofu: Remove this sentence ?

docs/howto/write_contract.md
```diff
@@ -20,3 +20,13 @@ pnpm add fhevm
...
+
+1/ For quick prototyping of a specific feature, use the [Zama version of the Remix IDE](./write_contract/remix.md). This will let you quickly deploy a contract on the devnet via Metamask, and interact easily with it throug the Remix UI. Otherwise, for a bigger project, you should start by cloning our custom [`fhevm-hardhat-template` repository](https://github.com/zama-ai/fhevm-hardhat-template). Hardhat is a popular development environment for Solidity developers and will let you test and deploy your contracts to the fhEVM using Javascript or Typescript code.
+
+2/ A good first step is to start with an unencrypted version of the contract you want to implement, as you would usually do on a regular EVM chain. It is easier to reason first on cleartext variables, before thinking on how to add confidentialy. You could test your contracts on a regular hardhat node via : `npx hardhat test --network hardhat`.
```
immortal-tofu:
```
A good first step is to start with an unencrypted version of the contract you want to implement, as you would usually do on a regular EVM chain. It is easier to reason first on cleartext variables, before thinking on how to add confidentialy.
```

docs/howto/write_contract/hardhat.md
```diff
@@ -4,6 +4,10 @@ The best way to start writing smart contracts with fhEVM is to use our [Hardhat
 
 It allows you to start a fhEVM docker image and run your smart contract on it. Read the [README](https://github.com/zama-ai/fhevm-hardhat-template/blob/main/README.md) for more information.
 
+When developing confidential contracts, we recommend to use first the mocked version of fhEVM for faster testing with `pnpm test:mock` and coverage computation via `pnpm coverage:mock`, this will lead to a better developer experience. But please keep in mind that the mocked fhEVM has some limitations and discrepencies compared to the real fhEVM node, as explained in the warning section at the end of this section.
```
immortal-tofu: `But please keep in mind` => `However, keep in mind`

docs/howto/write_contract/hardhat.md
```diff
@@ -4,6 +4,10 @@ The best way to start writing smart contracts with fhEVM is to use our [Hardhat
...
+
+Finally, always remember to test the final version of your contract on the real fhEVM with `pnpm test` before deployment.
```
immortal-tofu:
```
It's essential to run tests of the final contract version using the real fhEVM. You can do this by running `pnpm test` before deployment.
```

immortal-tofu (changes requested):

immortal-tofu:
Once merged in main, you need to open a pull request for `release/0.3.x` to make the changes happen in the current documentation.

jat9292 (approved):

immortal-tofu (approved):

PR #284, opened 2024-02-09
feat: add 64bits support

immortal-tofu:
- Tests are running on the CI launched manually with latest version of KMS/fhevm-go https://github.com/zama-ai/zbc-ethermint/actions/runs/7844274084/job/21408137996
- Mocks has been updated to support 64bits
- Contracts has been moved to euint64 (ERC20/Governor/Comp/Identity/...)
- fhevmjs has a prerelease which supports 64bits https://github.com/zama-ai/fhevmjs/releases


immortal-tofu:
Tests won't work until we have a proper dev image cc @leventdem

jatZama:
> Tests won't work until we have a proper dev image cc @leventdem

is this what is causing the nonce errors in the tests?

codegen/overloadTests.ts
```diff
@@ -50,7 +60,12 @@ export const overloadTests: { [methodName: string]: OverloadTest[] } = {
     { inputs: [0x00000001], output: 0xffffffff },
     { inputs: [0x00000002], output: 0xfffffffe },
   ],
-  not_euint32: [{ inputs: [0x0000fffe], output: 0xffff0001 }],
```
jatZama, line 53: not sure to understand this test output?

jatZama (commented):
correct not_euint32 in overload.ts

codegen/overloadTests.ts
immortal-tofu: This test didn't change in the last commit. You are probably looking at an old one :)

immortal-tofu:
> > Tests won't work until we have a proper dev image cc @leventdem
> 
> is this what is causing the nonce errors in the tests?

Yes, basically the first test to fail make the next tests failing for a nonce reason. Two things change regarding running tests:
- We need the new stack with 64bits support and tfhe-rs 0.5.1
- We need a computer capable to run multiplication over two euint64 in less than 7seconds: this is not possible on my M1 or on the default CI. That's why I switched to `large_ubuntu_16` in GitHub Actions.

david-zk (approved):
💪

codegen/overloadTests.ts
david-zk: @jatZama that is binary not, bit inversion, not boolean not

jatZama: I know this, here I was replying to a previous old commit version which has been corrected in a following commit.

jatZama (approved):

PR #292, opened 2024-02-22
feat: added the oracle contracts and library

jatZama:


oracle/OraclePredeploy.sol
```diff
@@ -0,0 +1,380 @@
...
+    }
+
+    modifier onlyRelayer() {
+        require(isRelayer[msg.sender], "Not relayer");
+        _;
+    }
+
+    // ONLY FOR TESTING PURPOSE : decryptTEST function MUST BE REMOVED IN PRODUCTION
+    // This is used in the test to simulate the threshold decryption by the KMS after the request
+    function decryptTEST(uint256 requestID) external view onlyRelayer returns (uint32) {
```
immortal-tofu: Can we find another way? Having a function to remove in production is error prone. Maybe it's possible to have a contract which extends this oracle for testing purpose?

immortal-tofu: Also I'd stick to camel case `decryptTest`

immortal-tofu (changes requested):

oracle/OraclePredeploy.sol
jatZama: The problem is that I do not know how to do otherwise, as this would change the hardcoded address of the predploy inside the library... Is there another way to do decryption without using TFHE.decrypt?

jatZama:
Second commit fixes all issues previously discussed here and on slack : 
-  added custom address in constructor for owner of predeploy
-  put the `decryptTest` (name in camelcase now) function into a separate contract `OraclePredeployTest` extending the original `OraclePredeploy`
- put the constant address of the Predeploy inside a separate `PredeployAddress.sol` file.

oracle/OraclePredeploy.sol
jatZama: answered and fixed here https://github.com/zama-ai/fhevm/pull/292#issuecomment-1964758613

immortal-tofu (approved):

PR #294, opened 2024-02-27
fix: silence warnings and fix _transfer in CompliantERC20

jatZama:


examples/GasEstimation.sol
```diff
@@ -6,279 +6,279 @@ import "../lib/TFHE.sol";
 contract GasEstimation {
     function empty() public {}
 
-    function randEuint8() public {
```
immortal-tofu, line 9: RandEuint8() can't be called as view method, since it writes something in state.

Add euint8 property to the contract to remove view.

immortal-tofu (approved):

PR #296, opened 2024-02-28
Feat/new types

immortal-tofu:
- Use native `ebool` from fhEVM instead of euint8
- Implement `euint4` type and operations

chore:
- Implement overload generator to create test with new types


examples/tests/TFHEManualTestSuite.sol
```diff
@@ -15,15 +15,15 @@ contract TFHEManualTestSuite {
         return TFHE.decrypt(TFHE.cmux(controlProc, ifTrueProc, ifFalseProc));
     }
 
-    function test_ebool_to_euint16_cast(bool input) public view returns (uint16) {
+    function test_eboolo_euint16_cast(bool input) public view returns (uint16) {
```
jatZama: typo here : eboolo -> ebool_to

examples/tests/TFHEManualTestSuite.sol
```diff
@@ -15,15 +15,15 @@ contract TFHEManualTestSuite {
...
         return TFHE.decrypt(TFHE.asEuint16(TFHE.asEbool(input)));
     }
 
-    function test_ebool_to_euint32_cast(bool input) public view returns (uint32) {
+    function test_eboolo_euint32_cast(bool input) public view returns (uint32) {
```
jatZama: maybe you can add casting to euint8, euint4 as well, and maybe from other types than ebool

codegen/overloadTests.ts
```diff
@@ -1,1297 +1,18 @@
+import overloads from './overloads.json';
 import { OverloadSignature, signatureContractMethodName } from './testgen';
 
 type OverloadTest = {
   inputs: number[];
```
jatZama, line 5: shouldn't inputs be also possibly be bigint[] tpye ?

jatZama: shouldn't inputs be also possibly be bigint[] tpye ?

codegen/generateOverloads.ts
```diff
@@ -0,0 +1,273 @@
...
+
+(BigInt as any).prototype['toJSON'] = function () {
+  return this.toString();
+};
+
+const SUPPORTED_UINT = [8, 16, 32, 64, 128, 256];
+const SUPPORTED_BITS = [4, 8, 16, 32, 64];
+
+const generateNumber = (bits: number) => {
+  return Math.max(Math.floor(Math.random() * (Math.pow(2, Math.min(bits, 28)) - 1)), 1);
```
jatZama: Maybe it would make sense to allow bigger numbers generated, up to 64 bits here for better testing, wdyt? so replacing 28 here by 64, and using bigin tinstead of number?

codegen/generateOverloads.ts
```diff
@@ -0,0 +1,273 @@
...
+      logs.push([lhsNumber, rhsNumber, result]);
+    }
+  }
+  return { inputs: [lhsNumber, rhsNumber], output: result };
+};
+
+export const SUPPORTED_FUNCTIONS: SupportedFunctions = {
+  add: {
+    supportedBits: SUPPORTED_BITS,
+    safeMin: true,
```
jatZama, line 79: not sure to understand why you would need the safeMin to be true here, if I understand this would mean that you force the output to be of type the lowest bitwidth of inputs, but this is not needed here, because you can indeed sum a euint8 with a euint64 and result would be a euint64

codegen/generateOverloads.ts
```diff
@@ -0,0 +1,273 @@
...
+          if (SUPPORTED_UINT.includes(lhs) && test.limit !== 'bits' && scalarCondition && !test.scalarOnly) {
+            lhsNumber = generateNumber(bitResults);
+            const encryptedTestName = [functionName, `uint${lhs}`, `euint${rhs}`].join('_');
+            const encryptedTests: Test[] = [];
+            if (!test.lhsHigher) {
+              encryptedTests.push(safeEval(test.evalTest, lhsNumber, rhsNumber, lhs, rhs, test.safeMin));
+              encryptedTests.push(safeEval(test.evalTest, smallest - 4, smallest, lhs, rhs, test.safeMin));
+            }
+            encryptedTests.push(safeEval(test.evalTest, smallest, smallest, lhs, rhs, test.safeMin));
+            encryptedTests.push(safeEval(test.evalTest, smallest, smallest - 4, lhs, rhs, test.safeMin));
```
jatZama: just noticed you were using safeEval for every test you pushed, but shouldnt we also test the wrapping behavior works correctly when overflowing? so also NOT using safeEval sometimes would make sense

codegen/overloadTests.ts
immortal-tofu: Yes potentially, for now I limit test with 32bits number, even for 64 bits. Especially because 64bits addition/multiplication ends with 128 bits integer.

codegen/generateOverloads.ts
immortal-tofu: I'll do that 👍

codegen/generateOverloads.ts
immortal-tofu: You're right, but the codegen expects a uint8 result, otherwise it fails. I think this is a mistake in the `testgen.ts` but I don't want to touch it now.

codegen/generateOverloads.ts
immortal-tofu: Overflowing check is already done on tfhe-rs lib, I don't think we need to go into this type of details.

examples/tests/TFHEManualTestSuite.sol
immortal-tofu: Good catch. Before ebool was a euint8, but it's not the case anymore. I'll add it.

codegen/generateOverloads.ts
jatZama: Ok, this is not so important indeed, but sometimes in rare cases we actually need to use the wrapping behavior , like what is done in the Uniswap V2 contract for the price oracle (only example I know of to be honest where overflow is desired) https://github.com/Uniswap/v2-core/blob/master/contracts/UniswapV2Pair.sol#L76

immortal-tofu: Let's test in the contract in this case.

codegen/overloads.json
```diff
@@ -1,2870 +1,2870 @@
...
+    { "inputs": ["4", "8"], "output": "12" },
+    { "inputs": ["8", "8"], "output": "16" },
+    { "inputs": ["8", "4"], "output": "12" }
   ],
   "add_euint64_euint32": [
-    { "inputs": [170118033, 266084851], "output": "436202884" },
-    { "inputs": [170118029, 170118033], "output": "340236062" },
-    { "inputs": [170118033, 170118033], "output": "340236066" },
-    { "inputs": [170118033, 170118029], "output": "340236062" }
+    { "inputs": ["17823", "2"], "output": "17825" },
```
jatZama: Maybe I am missing something but I was expecting some examples with real bigint outputs in the euint64 case... ?

immortal-tofu: Yes I just pushed a modification to change the strategy to generate random number.

codegen/generateOverloads.ts
jatZama: weird

jatZama (approved):

PR #299, opened 2024-03-01
test: cover the different types of values in decryption

youben11:


examples/TestAsyncDecrypt.sol
```diff
@@ -6,28 +6,93 @@ import "../lib/TFHE.sol";
...
+    }
+
+    function requestUint8() public {
+        euint8[] memory cts = new euint8[](1);
+        cts[0] = xUint8;
+        Oracle.requestDecryption(cts, this.callbackUint8.selector, 0, block.timestamp + 100);
+    }
+
+    function callbackUint8(uint256 requestID, uint8 decryptedInput) public onlyOracle returns (uint8) {
+        unchecked {
```
jatZama: you dont need to keep the unchecked block here as you dont do any arithmetic operation in your added examples. in my example i did an addition with two other numbers, and i added an unchecked because TFHE operators are wrapping in case of an overflow

jatZama: Also where are the Javascript tests???

examples/TestAsyncDecrypt.sol
```diff
@@ -6,28 +6,85 @@ import "../lib/TFHE.sol";
...
+        xUint64 = TFHE.asEuint64(64);
+    }
+
+    function requestBool() public {
+        ebool[] memory cts = new ebool[](1);
+        cts[0] = xBool;
+        Oracle.requestDecryption(cts, this.callbackBool.selector, 0, block.timestamp + 100);
+    }
+
+    function callbackBool(uint256 requestID, bool decryptedInput) public onlyOracle returns (bool) {
```
jatZama: Please replace `uint256 requestID` by `uint256 /*requestID*/` to mute compilation warnings during testing, as you are not using this argument in this case (contrarily to what I did for `callbackUint32`)

examples/TestAsyncDecrypt.sol
```diff
@@ -6,28 +6,85 @@ import "../lib/TFHE.sol";
...
+        return yBool;
+    }
+
+    function requestUint8() public {
+        euint8[] memory cts = new euint8[](1);
+        cts[0] = xUint8;
+        Oracle.requestDecryption(cts, this.callbackUint8.selector, 0, block.timestamp + 100);
+    }
+
+    function callbackUint8(uint256 requestID, uint8 decryptedInput) public onlyOracle returns (uint8) {
```
jatZama: Please replace `uint256 requestID` by `uint256 /*requestID*/` to mute compilation warnings during testing, as you are not using this argument in this case (contrarily to what I did for `callbackUint32`)

examples/TestAsyncDecrypt.sol
```diff
@@ -6,28 +6,85 @@ import "../lib/TFHE.sol";
...
     }
 
-    function request(uint32 input1, uint32 input2) public {
+    function requestUint16() public {
+        euint16[] memory cts = new euint16[](1);
+        cts[0] = xUint16;
+        Oracle.requestDecryption(cts, this.callbackUint16.selector, 0, block.timestamp + 100);
+    }
+
+    function callbackUint16(uint256 requestID, uint16 decryptedInput) public onlyOracle returns (uint16) {
```
jatZama: Please replace `uint256 requestID` by `uint256 /*requestID*/` to mute compilation warnings during testing, as you are not using this argument in this case (contrarily to what I did for `callbackUint32`)

examples/TestAsyncDecrypt.sol
```diff
@@ -6,28 +6,85 @@ import "../lib/TFHE.sol";
...
         }
     }
+
+    function requestUint64() public {
+        euint64[] memory cts = new euint64[](1);
+        cts[0] = xUint64;
+        Oracle.requestDecryption(cts, this.callbackUint64.selector, 0, block.timestamp + 100);
+    }
+
+    function callbackUint64(uint256 requestID, uint64 decryptedInput) public onlyOracle returns (uint64) {
```
jatZama: Please replace `uint256 requestID` by `uint256 /*requestID*/` to mute compilation warnings during testing, as you are not using this argument in this case (contrarily to what I did for `callbackUint32`)

test/oracleDecrypt/testAsyncDecrypt.ts
```diff
@@ -35,8 +35,8 @@ describe('TestAsyncDecrypt', function () {
       // An oracle was already deployed at ORACLE_PREDEPLOY_ADDRESS, check that is indeed the OraclePredploy contract by comparing hashes
       const codeHash = ethers.keccak256(codeAtAddress);
       if (
-        codeHash === '0x0cc63344ed0c9df080a56306bd363c6dc3618618cf3c2d77f0367c9fc745d99c' ||
-        codeHash === '0xdbf7e8e8bf5f776cb8b101f95ca1ff1ab161df3f9eb1f31cb39bfa66e9ba0034'
+        codeHash === '0x808db48789429e5f79adb9849400e0d01c5c0f017663f06c6b2ab48f2eeae10e' ||
```
jatZama: when I tested locally, the correct hash on fhevm is this one : `0xeffeaf5a1d024a603edb1c812f198ad8de84ea0a9b2b5e275b71a49794304d12` , so correct here please.
Second hash is correct and corresponds to the mocked mode. But I would prefer that you inverse order in the comment if you wish to keep this order in the hash. So finally I would write something like : 

     if (
        codeHash === '0xeffeaf5a1d024a603edb1c812f198ad8de84ea0a9b2b5e275b71a49794304d12' ||
        codeHash === '0xab7f73c011a5db013d92c7025cdc48b4d416ed21cd4e1ab102c590b46f560b26'
      ) {
        // it is indeed the OraclePredploy because it codeHashes are matching on either fhevm node or on the hardhat node (mocked mode)

jatZama (changes requested):

test/oracleDecrypt/testAsyncDecrypt.ts
youben11: working fine locally after the rebase on main

jatZama (approved):

PR #304, opened 2024-03-05
fix: updated mocked mode to be compatible with new types

jatZama:


codegen/templates.ts
```diff
@@ -626,10 +629,18 @@ function tfheCustomUnaryOperators(bits: number, signatures: OverloadSignature[])
         return euint${bits}.wrap(Impl.trivialEncrypt(value, Common.euint${bits}_t));
     }
 
+    `;
+  if (mocked) {
+    result += `
+    // Decrypts the encrypted 'value'.
+    function decrypt(euint${bits} value) internal view returns (${getUint(bits)}) {
+        return ${getUint(bits)}(Impl.decrypt(euint${bits}.unwrap(value)) % 2**${bits});
```
immortal-tofu, line 637: Why this part change?

codegen/templates.ts
```diff
@@ -753,6 +782,22 @@ function tfheCustomMethods(ctx: CodegenContext): string {
       return euint32.wrap(Impl.randBounded(upperBound, Common.euint32_t));
     }
 `;
+  if (mocked) {
+    result += `
+    // Decrypts the encrypted 'value'.
+    function decrypt(ebool value) internal view returns (bool) {
+        return (Impl.decrypt(ebool.unwrap(value)) % 2 == 1);
```
immortal-tofu, line 789: Why this part change?

immortal-tofu: Ho it's because you allow encryptBool to have any uint as value?

immortal-tofu (approved):

codegen/templates.ts
jatZama: because uint4 is not a native type in solidity which we rely on in mocked mode, we had some cases where this would be converted to a non-uint4 but a valid uint8 before this change.

codegen/templates.ts
jatZama: same reason as above, in native solidity bool are represented as uint8 under the hood, so some tests were failing before this change, now that we added ebool as a native type

.github/workflows/testmock.yml
```diff
@@ -0,0 +1,25 @@
+name: Pull request tests
```
immortal-tofu: `name: Pull request tests with mocks`

PR #313, opened 2024-03-08
refactor: remove deprecated optReq

immortal-tofu:


david-zk (approved):

codegen/main.ts
```diff
@@ -21,13 +21,6 @@ function generateAllFiles() {
     writeFileSync(`examples/tests/TFHETestSuite${os.shardNumber}.sol`, testgen.generateSmartContract(os));
   });
   writeFileSync('test/tfheOperations/tfheOperations.ts', testgen.generateTestCode(ovShards));
-  writeFileSync(
-    'test/generated.ts',
-    `
-    export const FHE_LIB_ADDRESS = '${context.libFheAddress}';
```
jatZama, line 27: why removing the FHE_LIB_ADDRESS? i guess it is not needed?

jatZama (approved):

codegen/main.ts
immortal-tofu: I've introduced `getPublicKeyCallParams` in fhevmjs, which provide address and data directly. https://github.com/zama-ai/fhevm/blob/main/test/instance.ts#L42

PR #314, opened 2024-03-08
feat: introduce select in place of cmux

immortal-tofu:


jatZama:
Can you run the codegen before merging? there is still no select in the TFHE library. Also please update the tests.

jatZama (commented):
Can you run the codegen before merging?

jatZama (commented):
Ok for me but the async decrypt oracle test might fail probably because the codehash of the predploy would have changed.

immortal-tofu:
> Ok for me but the async decrypt oracle test might fail probably because the codehash of the predploy would have changed.

We'll need to update it

jatZama (approved):

PR #321, opened 2024-03-11
docs: various adjustments to the current documentation

immortal-tofu:


docs/fundamentals/types.md
```diff
@@ -12,52 +12,46 @@ The `e(u)int` types are **wrappers** over these handles.
...
   totalSupply = totalSupply + amount;
 }
 ```
 
 ## Contract state variables with encrypted types
 
-If you require a state variable that utilizes these encrypted types, you cannot directly assign the value. In Solidity, the compiler attempts to ascertain the value of `TFHE.asEuintXX(yy)` during compilation, which is not feasible because `asEuintXX()` invokes a precompiled contract. To address this challenge, you should declare your property and subsequently assign its value within the constructor.
-Also, please be aware that you should never declare encrypted types as either constant or immutable variables, as these variables need to be stored in the privileged storage of the fhEVM, which is not compatible with constant variable types.
-
-```
```
jatZama, line 57: Maybe keep this example?

immortal-tofu: I wanted to remove it because, AFAIU, you could write:
```
euint32 private totalSupply = TFHE.asEuint32(0);
```
Just `immutable` and `constant` are problematic, so that's why I rewrite with this idea in mind. Can you confirm?

jatZama: yes just those 2 are problematic, and you could use either `euint32 private totalSupply = TFHE.asEuint32(0);` during state variable declaration or inside the constructor. Alternatively put both examples perhaps.

immortal-tofu: So you want me to add 
```
euint32 private totalSupply = TFHE.asEuint32(0);
```
```
euint32 private totalSupply;
constructor() {
  totalSupply = TFHE.asEuint32(0);
}
```

It doesn't seem very useful regarding the constant/immutable limitation, no?

jatZama: Yes, I thought it would make things clearer, but if you think it is not you can remove the examples.

jatZama (approved):

PR #323, opened 2024-03-12
docs: added docs for async decrypt and support for euint4

jatZama:


docs/fundamentals/decrypt.md
```diff
@@ -6,6 +6,141 @@ We allow explicit decryption requests for any encrypted type. The values are dec
...
+    uint256[] memory params = getParamsUint(requestID);
+    unchecked {
+        uint32 result = uint32(params[0]) + uint32(params[1]) + decryptedInput;
+        yUint32 = result;
+        return result;
+    }
+}
+```
+
+### Old example (deprecated)
```
immortal-tofu: Can you remove the documentation regarding sync decrypt? We discussed it with Rand and he agrees that we shouldn't have documentation on deprecated feature.

docs/fundamentals/decrypt.md
```diff
@@ -6,6 +6,141 @@ We allow explicit decryption requests for any encrypted type. The values are dec
...
+  }
+
+  function myCustomCallback(uint256 /*requestID*/, bool decryptedInput) public onlyOracle returns (bool) {
+    yBool = decryptedInput;
+    return yBool;
+  }
+```
+
+> **_WARNING:_**
+> For the previous code to work properly, please ensure that the `OraclePredeploy` contract was already deployed to the expected address `ORACLE_PREDEPLOY_ADDRESS` hardcoded inside [`fhevm/oracle/lib/PredeployAddress.sol`](../../oracle/lib/PredeployAddress.sol)! To correctly deploy `OraclePredeploy`, it must be done as Eve's first transaction, i.e when here seed is null, and where Eve's private key is defined as `0x717fd99986df414889fd8b51069d4f90a50af72e542c58ee065f5883779099c6`. Furthermore, the predeploy owner should add the authorized relayers addresses by calling the `addRelayer` function on `OraclePredploy`. Relayers are the only accounts authorized to fulfil the decryption requests. However `OraclePredploy` would still check the KMS signature during the fulfilment, so we trust the relayer only to forward the request on time, a rogue relayer could not cheat by sending fake decryption results.
```
immortal-tofu: We shouldn't document that: a developer doesn't have to deploy the Oracle. This documentation should be present on the integration documentation (here https://github.com/zama-ai/fhevm-go/tree/main/docs)

docs/fundamentals/decrypt.md
jatZama: Should I remove this subsection entirely?
Maybe I should keep part of the last warnings from the last part ```For now, a `TFHE.decrypt` is pretty cheap, making it tempting to use constructs like `if(TFHE.decrypt(encryptedBool))`. However, it is recommended to avoid this approach, as in the future, each decryption will trigger an external call, introducing latency and incurring gas costs. Instead, use [select operator to handle conditions](conditions.md).
For the same reason, you should replace a `require(TFHE.decrypt(encryptedBool1) && TFHE.decrypt(encryptedBool2));` with a `TFHE.decrypt(TFHE.and(encryptedBool1, encryptedBool2));` to limit decryption calls.```

docs/fundamentals/decrypt.md
jatZama: Ok I was not aware of the existence of this documentation, I did not see it on Zama website. Should I add a PR on this repo as well?

docs/fundamentals/decrypt.md
immortal-tofu: Remove it entirely. We don't want to promote the fact a sync decrypt still exists. We just keep it in fhEVM for old contract in the meantime.

docs/fundamentals/decrypt.md
immortal-tofu: It would be great. There is a link in the README of fhevm-go, but it's not listed on the menu https://docs.zama.ai/fhevm-go
One objective for next quarter is to clean it and add more documentation.

jatZama: To be honest, I think that @youben11 is more able than me to add documentation for fhevm-go , as he knows the codebase better than me.

immortal-tofu: Juste remove this part in this case, we'll add this in fhevm-go later

docs/fundamentals/decrypt.md
```diff
@@ -6,21 +6,146 @@ We allow explicit decryption requests for any encrypted type. The values are dec
...
+    unchecked {
+        uint32 result = uint32(params[0]) + uint32(params[1]) + decryptedInput;
+        yUint32 = result;
+        return result;
+    }
 }
 ```
 
-For now, a `TFHE.decrypt` is pretty cheap, making it tempting to use constructs like `if(TFHE.decrypt(encryptedBool))`. However, it is recommended to avoid this approach, as in the future, each decryption will trigger an external call, introducing latency and incurring gas costs. Instead, use [select operator to handle conditions](conditions.md).
+When the decryption request is fufilled by the relayer, the `OraclePredploy` contract, when calling the callback function, will also emit one of the following events, depending on the type of requested ciphertext:
```
immortal-tofu: the `OraclePredeploy` contract

immortal-tofu (approved):

PR #329, opened 2024-03-14
feat: added automatic deployment of predeploy when testing

jatZama:


package.json
```diff
@@ -27,7 +27,7 @@
     "task:deployOracle": "hardhat task:deployOracle",
     "task:addRelayer": "hardhat task:addRelayer",
     "task:removeRelayer": "hardhat task:removeRelayer",
-    "fhevm:start": "docker run -i -p 8545:8545 --rm --name fhevm ghcr.io/zama-ai/ethermint-dev-node:v0.3.0-5",
+    "fhevm:start": "chmod +x launch-fhevm.sh && ./launch-fhevm.sh",
```
youben11: why run `chmod` on every call? You can push it in github as executable

immortal-tofu (approved):

package.json
jat9292: Do you mean if i do `chmod` locally then I push it would work?

jatZama: Do you mean if i do `chmod` locally then I push it would work?

youben11: yes

test/asyncDecrypt.ts
```diff
@@ -0,0 +1,317 @@
...
+const ifaceResultCallbackUint32 = new ethers.Interface(['event ResultCallbackUint32' + argEvents2]);
+const ifaceResultCallbackUint64 = new ethers.Interface(['event ResultCallbackUint64' + argEvents2]);
+
+let oracle: OraclePredeploy;
+let firstBlockListening: number;
+
+export const asyncDecrypt = async (): Promise<void> => {
+  firstBlockListening = await ethers.provider.getBlockNumber();
+  // this function will emit logs for every request and fulfilment of a decryption
+  oracle = await ethers.getContractAt('OraclePredeploy', parsedEnv.ORACLE_CONTRAT_PREDEPLOY_ADDRESS);
```
david-zk: Typo?
`ORACLE_CONTRAT_PREDEPLOY_ADDRESS`
should be
`ORACLE_CONTRACT_PREDEPLOY_ADDRESS`

jatZama: good catch, this was badly written dozens of times. I pushed a fix.

package.json
jatZama: done

immortal-tofu (approved):

PR #348, opened 2024-04-02
docs: update roadmap

immortal-tofu:


docs/fundamentals/conditions.md
```diff
@@ -1,4 +1,4 @@
-# Write conditions
+# Branching in FHE
```
aquint-zama, line 1: wonder if we should say somewhere that `select` is former `CMUX`?

aquint-zama (approved):

PR #349, opened 2024-04-02
feat: add ability to call eaddress decrypt asynchronously

leventdem:


immortal-tofu (approved):

codegen/templates.ts
```diff
@@ -21,6 +22,8 @@ library Common {
     uint8 internal constant euint16_t = 3;
     uint8 internal constant euint32_t = 4;
     uint8 internal constant euint64_t = 5;
+    uint8 internal constant euint128_t = 6;
+    uint8 internal constant euint160_t = 7;
```
jatZama, line 26: can you change this to eaddress please?

examples/TestAsyncDecrypt.sol
```diff
@@ -100,4 +103,15 @@ contract TestAsyncDecrypt is OracleCaller {
...
         return decryptedInput;
     }
+
+    function requestAddress() public {
+        eaddress[] memory cts = new eaddress[](1);
+        cts[0] = xAddress;
+        Oracle.requestDecryption(cts, this.callbackAddress.selector, 0, block.timestamp + 100);
+    }
+
+    function callbackAddress(uint256 /*requestID*/, uint160 decryptedInput) public onlyOracle returns (address) {
```
jatZama, line 113: pleae change this to address instead of uint160

examples/TestAsyncDecrypt.sol
```diff
@@ -27,6 +29,7 @@ contract TestAsyncDecrypt is OracleCaller {
         xUint16 = TFHE.asEuint16(16);
         xUint32 = TFHE.asEuint32(32);
         xUint64 = TFHE.asEuint64(64);
+        xAddress = TFHE.asEaddress(797161134358056856230896843146392277790002887282);
```
jatZama, line 32: input should be an address instead of uint160

mocks/TFHE.sol
```diff
@@ -17,6 +18,8 @@ library Common {
     uint8 internal constant euint16_t = 3;
     uint8 internal constant euint32_t = 4;
     uint8 internal constant euint64_t = 5;
+    uint8 internal constant euint128_t = 6;
+    uint8 internal constant euint160_t = 7;
```
jatZama, line 22: can you change this to eaddress please?

oracle/OraclePredeploy.sol
```diff
@@ -416,6 +460,26 @@ contract OraclePredeploy is Ownable2Step {
         isFulfilled[requestID] = true;
     }
 
+    function fulfillRequestAddress(
+        uint256 requestID,
+        uint160[] memory decryptedCt /*, bytes memory signatureKMS */
```
jatZama, line 465: should be an array of addresses instead of uint160s

jatZama (changes requested):

oracle/OraclePredeploy.sol
immortal-tofu: Under the hood, everything is uint160. I think we should keep everything as uint160, but when we call the callback function, we should pass an array of address. For this, we should convert the received uint160 from the relayer to actual address.

Maybe something like `callbackCalldata = abi.encodePacked(callbackCalldata, abi.encode(address(uint160(decryptedCt[i]))));` would work?

examples/TestAsyncDecrypt.sol
immortal-tofu: Both should work. Probably we need :

```
    // Convert an address value to an encrypted asEaddress.
    function asEaddress(address value) internal pure returns (eaddress) {
        return eaddress.wrap(Impl.trivialEncrypt(uint160(value), Common.euint160_t));
    }
```

codegen/templates.ts
immortal-tofu: euint160_t is the type used in fhevm-go, in tfhe-rs and in the KMS. I'd prefer keep euint160_t.

immortal-tofu:
User shouldn't manipulate numbers when we talk about eaddress or decryption of eaddress. User needs these functions:
```
TFHE.asEaddress(0x028783f98...);

function myCallback(uint256 requestId, address[] memory decryptions) {
  ...
}
```
If it's missing we should add this. Under the hood, tfhe-rs/KMS/fhevm-go talk about uint160 but it's fine. We just do the conversion in TFHE.sol on the fly with `address(value)` or `uint160(value)`

Note: I'm off this week, so I let Levent decide what to do, especially regarding the release.

mocks/TFHE.sol
immortal-tofu: This value is linked to the implementation in fhevm-go, where we only talk about uint160. I'd prefer keep uint160_t.

PR #354, opened 2024-04-09
Fix/euint160 nooracleservicechange

jatZama:


leventdem (approved):
Great to be consistent with e(address) everywhere +1

PR #358, opened 2024-04-11
feat: new async-decrypt API

jatZama:
This new API is cleaner and more concise (no more almost duplicated events for each types of ciphertexts) and allows users to request decryption of different types in a single `requestDecryption` transaction.
Work in Progress status: DONE
1st commit ready : all the solidity code is done and ready for review already ✅
2nd commit will come soon for the JS part (tests+mocked mode). ✅ NOW DONE! 

For the solidity part, discussion and approval needed on this:
I had to decouple the storage of honestly obtained ciphertext in privileged storage - via the `OraclePredeploy:approveEUintXXX` functions, called inside the `Oracle:toCiphertext` internal library functions -  from the `requestDecryption` in order to be able to send ciphertexts of different types in a single `requestDecryption`call. This will add some complexity in the oracle service as we will now have to loop on all stored array of handles in the `verifiedEUintXXXs` mappings (for a fixed counter/key index and type) until we find at least one handle which comes from a honestly obtained ciphertext, i.e with corresponding ciphertext in privileged storage. Also security model guarantees are not 100% the same - in theory a user could approve a honestly obtained handle in a first transaction and (some other malicious user) request decryption in a later transaction even with non-honestly obtained identical handle if counter did not change i.e if no `requestDecryption` happened meanwhile- but risk is inexistent as long as user ensures there is no call to an `approveEUintXXX` function unless he wanted to request a decryption in the same transaction. In summary there should be no risk unless the smart contract of the dApp is malicious.

Also, it is now the responsibility of the relayer to do the correct abi-encoding of the concatenated decrypted variables and pass them as a single bytes array argument to the `fulfillRequest` function of the `OraclePredeploy`

### DIFF
--- a/examples/TestAsyncDecrypt.sol
+++ b/examples/TestAsyncDecrypt.sol
@@ -208,6 +208,37 @@ contract TestAsyncDecrypt is OracleCaller {
     function callbackAddress(uint256, address decryptedInput) public onlyOracle returns (address) {
         yAddress = decryptedInput;
         return decryptedInput;
+
+    function requestMixed() public {
+        Ciphertext[] memory cts = new Ciphertext[](10);
+        cts[0] = Oracle.toCiphertext(xBool);
+        cts[1] = Oracle.toCiphertext(xBool);
+        cts[2] = Oracle.toCiphertext(xUint4);
+        cts[3] = Oracle.toCiphertext(xUint8);
+        cts[4] = Oracle.toCiphertext(xUint16);
+        cts[5] = Oracle.toCiphertext(xUint32);
+        cts[6] = Oracle.toCiphertext(xUint64);
+        cts[7] = Oracle.toCiphertext(xUint64);
+        cts[8] = Oracle.toCiphertext(xUint64);
+        cts[9] = Oracle.toCiphertext(xAddress);
+        Oracle.requestDecryption(cts, this.callbackMixed.selector, 0, block.timestamp + 100);
+    }
+
+    function callbackMixed(
+        uint256 /*requestID*/,
+        bool decBool_1,
+        bool decBool_2,
+        uint8 decUint4,
+        uint8 decUint8,
+        uint16 decUint16,
+        uint32 decUint32,
+        uint64 decUint64_1,
+        uint64 decUint64_2,
+        uint64 decUint64_3,
+        address decAddress
+    ) public onlyOracle returns (uint8) {
+        yUint4 = decUint4;
+        return yUint4;
     }
 
     function requestMixed(uint32 input1, uint32 input2) public {
--- a/examples/TestAsyncDecrypt.sol
+++ b/examples/TestAsyncDecrypt.sol
@@ -250,5 +250,4 @@ contract TestAsyncDecrypt is OracleCaller {
         yAddress = decAddress;
         return yUint4;
     }
-
 }
--- a/examples/TestAsyncDecrypt.sol
+++ b/examples/TestAsyncDecrypt.sol
@@ -56,6 +56,21 @@ contract TestAsyncDecrypt is OracleCaller {
         return yBool;
     }
 
+    function requestBoolInfinite() public {
+        ebool[] memory cts = new ebool[](1);
+        cts[0] = xBool;
+        Oracle.requestDecryption(cts, this.callbackBoolInfinite.selector, 0, block.timestamp + 100);
+    }
+
+    function callbackBoolInfinite(uint256 /*requestID*/, bool decryptedInput) public onlyOracle returns (bool) {
+        uint256 i = 0;
+        while (1 == 1) {
+            i++;
+        }
+        yBool = decryptedInput;
+        return yBool;
+    }
+
     function requestBoolAboveDelay() public {
         // should revert
         Ciphertext[] memory cts = new Ciphertext[](1);
--- a/examples/TestAsyncDecrypt.sol
+++ b/examples/TestAsyncDecrypt.sol
@@ -210,7 +210,7 @@ contract TestAsyncDecrypt is OracleCaller {
         return decryptedInput;
     }
 
-    function requestMixed() public {
+    function requestMixed(uint32 input1, uint32 input2) public {
         Ciphertext[] memory cts = new Ciphertext[](10);
         cts[0] = Oracle.toCiphertext(xBool);
         cts[1] = Oracle.toCiphertext(xBool);
@@ -222,11 +222,13 @@ contract TestAsyncDecrypt is OracleCaller {
         cts[7] = Oracle.toCiphertext(xUint64);
         cts[8] = Oracle.toCiphertext(xUint64);
         cts[9] = Oracle.toCiphertext(xAddress);
-        Oracle.requestDecryption(cts, this.callbackMixed.selector, 0, block.timestamp + 100);
+        uint256 requestID = Oracle.requestDecryption(cts, this.callbackMixed.selector, 0, block.timestamp + 100);
+        addParamsUint(requestID, input1);
+        addParamsUint(requestID, input2);
     }
 
     function callbackMixed(
-        uint256,
+        uint256 requestID,
         bool decBool_1,
         bool decBool_2,
         uint8 decUint4,
@@ -238,7 +240,19 @@ contract TestAsyncDecrypt is OracleCaller {
         uint64 decUint64_3,
         address decAddress
     ) public onlyOracle returns (uint8) {
+        yBool = decBool_1;
+        require(decBool_1 == decBool_2, "Wrong decryption");
         yUint4 = decUint4;
+        yUint8 = decUint8;
+        yUint16 = decUint16;
+        uint256[] memory params = getParamsUint(requestID);
+        unchecked {
+            uint32 result = uint32(params[0]) + uint32(params[1]) + decUint32;
+            yUint32 = result;
+        }
+        yUint64 = decUint64_1;
+        require(decUint64_1 == decUint64_2 && decUint64_2 == decUint64_3, "Wrong decryption");
+        yAddress = decAddress;
         return yUint4;
     }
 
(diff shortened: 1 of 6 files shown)


examples/TestAsyncDecrypt.sol
```diff
@@ -50,75 +50,35 @@ contract TestAsyncDecrypt is OracleCaller {
...
+        cts[0] = Oracle.toCiphertext(xBool);
+        cts[1] = Oracle.toCiphertext(xBool);
+        cts[2] = Oracle.toCiphertext(xUint4);
+        cts[3] = Oracle.toCiphertext(xUint8);
+        cts[4] = Oracle.toCiphertext(xUint16);
+        cts[5] = Oracle.toCiphertext(xUint32);
+        cts[6] = Oracle.toCiphertext(xUint64);
+        cts[7] = Oracle.toCiphertext(xUint64);
+        cts[8] = Oracle.toCiphertext(xUint64);
+        cts[9] = Oracle.toCiphertext(xAddress);
```
david-zk, line 219: Why is this long array needed? Seems like reflection on previous multiple events per type.
Is it possible to do without it?

jatZama: This is an example **test** file where I mix every kind of types, some with duplication. We could never do this before, because the Oracle only supported an array containing a single type.

david-zk: Oh, my bad this is test

david-zk:
@jatZama Also I'm curious, maybe the idea to bind to ciphertexts is not ideal, like, what if we add another type, we can't redeploy smart contract. Maybe the ciphertext type should be just a number, I'm curious

jatZama:
> @jatZama Also I'm curious, maybe the idea to bind to ciphertexts is not ideal, like, what if we add another type, we can't redeploy smart contract. Maybe the ciphertext type should be just a number, I'm curious

Not sure to understand exactly what do you mean. Adding new types in the future should be more straigthforward, as you only need to update the `CiphertextType` enum, you can already see btw that I prepared support for EUINT128 in the future despite not being implemented by fhevm yet https://github.com/zama-ai/fhevm/blob/1fc5feda7fd4d448c3dd345aacd271ccd17ea908/oracle/OraclePredeploy.sol#L8

david-zk:
> > @jatZama Also I'm curious, maybe the idea to bind to ciphertexts is not ideal, like, what if we add another type, we can't redeploy smart contract. Maybe the ciphertext type should be just a number, I'm curious
> 
> Not sure to understand exactly what do you mean. Adding new types in the future should be more straigthforward, as you only need to update the `CiphertextType` enum, you can already see btw that I prepared support for EUINT128 in the future despite not being implemented by fhevm yet
> 
> https://github.com/zama-ai/fhevm/blob/1fc5feda7fd4d448c3dd345aacd271ccd17ea908/oracle/OraclePredeploy.sol#L8

Yeah but these variables
```
    mapping(uint256 => ebool[]) internal verifiedEBools; // to check that ciphertexts have been honestly obtained via privileged storage
    mapping(uint256 => euint4[]) internal verifiedEUint4s;
    mapping(uint256 => euint8[]) internal verifiedEUint8s;
    mapping(uint256 => euint16[]) internal verifiedEUint16s;
    mapping(uint256 => euint32[]) internal verifiedEUint32s;
    mapping(uint256 => euint64[]) internal verifiedEUint64s;
    mapping(uint256 => eaddress[]) internal verifiedEAddresses;
```
Correct me if I'm wrong but I believe we can't redeploy smart contract with new variables?

oracle/OraclePredeploy.sol
```diff
@@ -5,138 +5,52 @@ pragma solidity ^0.8.20;
...
-        uint256 msgValue,
-        uint256 maxTimestamp
-    );
+    mapping(uint256 => ebool[]) internal verifiedEBools; // to check that ciphertexts have been honestly obtained via privileged storage
+    mapping(uint256 => euint4[]) internal verifiedEUint4s;
+    mapping(uint256 => euint8[]) internal verifiedEUint8s;
+    mapping(uint256 => euint16[]) internal verifiedEUint16s;
+    mapping(uint256 => euint32[]) internal verifiedEUint32s;
+    mapping(uint256 => euint64[]) internal verifiedEUint64s;
+    mapping(uint256 => eaddress[]) internal verifiedEAddresses;
```
dartdart26: Maybe a dumb question, but I didn't fully grasp why we need these mappings. Can you explain a bit more?

dartdart26:
API looks good! I just have one question that I posted above.

oracle/OraclePredeploy.sol
jatZama: Because we need to store the honestly obtained ciphertexts in privileged storage of the OraclePredeploy. I am doing this inside the Oracle library, it is abstraced away when calling the toCiphertext library function https://github.com/zama-ai/fhevm/blob/1fc5feda7fd4d448c3dd345aacd271ccd17ea908/oracle/lib/Oracle.sol#L15-L16
Is there another way to do this? 
Btw in next commit I will also add the no-op inside each `OraclePredeploy:approveEuintXXX` function just before storing the ciphertext in the mapping.

jatZama:
> > > @jatZama Also I'm curious, maybe the idea to bind to ciphertexts is not ideal, like, what if we add another type, we can't redeploy smart contract. Maybe the ciphertext type should be just a number, I'm curious
> > 
> > 
> > Not sure to understand exactly what do you mean. Adding new types in the future should be more straigthforward, as you only need to update the `CiphertextType` enum, you can already see btw that I prepared support for EUINT128 in the future despite not being implemented by fhevm yet
> > https://github.com/zama-ai/fhevm/blob/1fc5feda7fd4d448c3dd345aacd271ccd17ea908/oracle/OraclePredeploy.sol#L8
> 
> Yeah but these variables
> 
> ```
>     mapping(uint256 => ebool[]) internal verifiedEBools; // to check that ciphertexts have been honestly obtained via privileged storage
>     mapping(uint256 => euint4[]) internal verifiedEUint4s;
>     mapping(uint256 => euint8[]) internal verifiedEUint8s;
>     mapping(uint256 => euint16[]) internal verifiedEUint16s;
>     mapping(uint256 => euint32[]) internal verifiedEUint32s;
>     mapping(uint256 => euint64[]) internal verifiedEUint64s;
>     mapping(uint256 => eaddress[]) internal verifiedEAddresses;
> ```
> 
> Correct me if I'm wrong but I believe we can't redeploy smart contract with new variables?

The issue is that we still need to store honestly obtained ciphertexts inside the privileged storage of `OraclePredeploy`. 
If we need to add other mappings in the future we could make the OraclePredeploy contract upgradable, using UUPS pattern, but I am not sure this is a good solution security-wise, and this was never discussed before. Wdyt @dartdart26 ?

oracle/OraclePredeploy.sol
dartdart26: Wouldn't the handle eventually be stored in OraclePredeploy via the `requestDecryption()` call?

jatZama: How would you do it? 
Inside requestDecryption it is no longer possible because we pass the new Ciphertext type which is a struct {uint255, CipetextType}
unless i am missing something

dartdart26: Putting here for reference for others. As discussed offline, wrapping won't prohibit the fhEVM from seeing inside the `Ciphertext` data type and treating the uint256 as a handle. Then, SSTORE would eventually also treat it as a handle and persist the ciphertext.

dartdart26: So we can use the `decryptionRequests` mapping only.

jat9292:
> This new API is cleaner and more concise (no more almost duplicated events for each types of ciphertexts) and allows users to request decryption of different types in a single `requestDecryption` transaction. Work in Progress status: 1st commit ready : all the solidity code is done and ready for review already ✅ 2nd commit will come soon for the JS part (tests+mocked mode). 🚧
> 
> For the solidity part, discussion and approval needed on this: I had to decouple the storage of honestly obtained ciphertext in privileged storage - via the `OraclePredeploy:approveEUintXXX` functions, called inside the `Oracle:toCiphertext` internal library functions - from the `requestDecryption` in order to be able to send ciphertexts of different types in a single `requestDecryption`call. This will add some complexity in the oracle service as we will now have to loop on all stored array of handles in the `verifiedEUintXXXs` mappings (for a fixed counter/key index and type) until we find at least one handle which comes from a honestly obtained ciphertext, i.e with corresponding ciphertext in privileged storage. Also security model guarantees are not 100% the same - in theory a user could approve a honestly obtained handle in a first transaction and (some other malicious user) request decryption in a later transaction even with non-honestly obtained identical handle if counter did not change i.e if no `requestDecryption` happened meanwhile- but risk is inexistent as long as user ensures there is no call to an `approveEUintXXX` function unless he wanted to request a decryption in the same transaction. In summary there should be no risk unless the smart contract of the dApp is malicious.
> 
> Also, it is now the responsibility of the relayer to do the correct abi-encoding of the concatenated decrypted variables and pass them as a single bytes array argument to the `fulfillRequest` function of the `OraclePredeploy`

Just realized thanks to @dartdart26 that ciphertext would still get verified even after unwrapping them to uint256, which makes sense since custom types are not native. This allowed me to simplify quite a bit the PR, by removing all the verifiedCiphertexts mappings , and also removing some risk because now we no more have decoupling between ciphertext verification and decryption request, everything happens during the request. Please check new commit, btw it also solves the concern @david-zama raised about lack of upgradability.
 The only question still pending is : should I remove the `CiphertextType` enum? It is better to keep it for readability for the moment at least I think, also I am using it for the no-op logic during the request. Maybe in the future we could remove it if we decide to implement a standard no-op operator common for all encrypted types. For instance for the moment, `eaddress` supports only operators which are not supported by ebool, so I had to do use a different no-op in this case.

jatZama:
Mocked mode is now ready. Only remaining work now is on oracle service side.

leventdem (commented):
Looks great, very useful refactoring that allows multiple types to be decrypted at once  and a cleaner code. Moreover, handling the Ciphertext as a struct of payload and FHE type is the way used everywhere after the oracle (aggregator, kms). 

Let's wait the oracle update to make the CI happy and hopefully merge.

jatZama (commented):
LGTM

leventdem (approved):

PR #361, opened 2024-04-15
chore: bug return bool

jatZama:


jatZama:
To reproduce the bug simply run : `npm run test test/testEq/*`

jatZama:
My bad, I was missing `await contract2.waitForDeployment();` after deployment. I will close this PR .

PR #373, opened 2024-04-23
docs: local node must import OracleCaller

jatZama:


docs/getting_started/local.md
```diff
@@ -17,6 +17,11 @@ docker run -i -p 8545:8545 -p 8546:8546 --rm --name fhevm ghcr.io/zama-ai/etherm
 
 <!-- markdown-link-check-enable -->
 
+However, we advise developers to use directly `pnpm fhevm:start` or `npm run fhevm:start` commands instead of the previous command, as this will launch a bash script which will also deploy automatically the oracle contract and launch the oracle relayer service, which are needed for asynchornous decryption requests.
```
immortal-tofu: `asynchornous` => `asynchronous`

We should also mention that these commands are available in the hardhat template, otherwise it doesn't make sense without the context

docs/getting_started/local.md
```diff
@@ -17,6 +17,11 @@ docker run -i -p 8545:8545 -p 8546:8546 --rm --name fhevm ghcr.io/zama-ai/etherm
...
+
+> **_WARNING:_** > `OracleCaller.sol` must be imported at least once in one of your smart contracts if you wish to use the recommended `fhevm:start` command, or else the bash script will emit an error and decryptions would fail. This is needed because hardhat needs to compile the oracle predeploy contract before your initial deployment. This can be done simply by adding the following import at the top of any of the smart contracts used in your project:
```
immortal-tofu, line 22: We can avoid this if we publish the compiled version in the fhevm package, no?

immortal-tofu (changes requested):

docs/getting_started/local.md
jatZama: I am not sure but I don't believe it would work, as the artifcats json files must be located inside the artifacts/ folder resulting from compilation of source contracts.

immortal-tofu: You're right. Too bad :(

immortal-tofu (approved):

PR #380, opened 2024-05-07
Add regression

david-zk:
Adds regression test

@jatZama

### DIFF
--- a/examples/regressions/Regression1.sol
+++ b/examples/regressions/Regression1.sol
@@ -1,0 +1,41 @@
+/* SPDX-License-Identifier: MIT */
+pragma solidity ^0.8.20;
+
+contract Regression1 {
+    error IndexOutOfBound();
+    struct Metadata {
+        uint256 created;
+        uint256 lastUpdated;
+        uint256 versionId;
+    }
+    struct Service {
+        bytes32 id;
+        string serviceType;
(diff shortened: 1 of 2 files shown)


immortal-tofu:
Is it still relevant @david-zk ?

jatZama:
This was an example of transaction without any FHE computation (from an issue found by Shiba, but I also found similar issues few times) which failed on fhevm due to wrong gas estimation. @dartdart26 asked me about it few days ago, something to keep in mind because this kind of failure is not expected at all.

immortal-tofu:
Can we close this PR @jatZama ?

jatZama:
You can close if you want, since this Regression test is already present here, so we can keep track of it since ethermint still has issues in gas estimation: https://github.com/zama-ai/fhevm/blob/main/examples/Regression1.sol
For the moment it is "solved" by overriding gasLimit by +20% of original value in the extended provider here https://github.com/zama-ai/fhevm/blob/main/CustomProvider.ts#L19 
Note that this trick does not solve the gas estimation error we get in the async decrypt tests however, so in those we are still using a hardcoded gasLimit in ethermint as you can see here: https://github.com/zama-ai/fhevm/blob/main/test/gatewayDecrypt/testAsyncDecrypt.ts#L22

PR #385, opened 2024-05-23
Feat: add array equality

immortal-tofu:


mocks/Impl.sol
```diff
@@ -65,6 +65,16 @@ library Impl {
         result = (lhs == rhs) ? 1 : 0;
     }
 
+    function eq(uint256[] memory larray, uint256[] memory rarray) internal pure returns (uint256 result) {
+        if (larray.length != rarray.length) {
+            return false;
+        }
+        result = true;
+        for (uint i = 0; i < larray.length; i++) {
+            if (larray[i] != rarray[i]) result = false;
```
jatZama: Nit: you can return; at the end of this line to save some gas

mocks/TFHE.sol
```diff
@@ -5262,6 +5262,66 @@ library TFHE {
         return euint64.wrap(Impl.select(ebool.unwrap(control), euint64.unwrap(a), euint64.unwrap(b)));
     }
 
+    function eq(euint4[] memory a, euint4[] memory b) internal pure returns (ebool) {
```
jatZama, line 5265: are we sure we don't want to support eq for ebool[] and eaddress[] as well? native eq for ebool is not supported but this is equivalent to not(xor(a,b))

mocks/TFHE.sol
```diff
@@ -5262,6 +5262,66 @@ library TFHE {
...
+        uint256[] memory larray;
+        uint256[] memory rarray;
+        for (uint i = 0; i < a.length; i++) {
```
jatZama, line 5269: I think you need to add a require to ensure a.length is equal to b.length, because I assume the tx reverts in FHEVMLib precompile if lengths are unequal (to be checked).

jatZama (changes requested):
See my comments concerning missing require on length, early return and missing types.
In addition, adding some tests will be nice.

mocks/TFHE.sol
immortal-tofu: This is not implemented in tfhe-rs for 160bits. I didn't check for ebool, I'll take a look.

immortal-tofu:
> See my comments concerning missing require on length, early return and missing types. In addition, adding some tests will be nice.

I'll add some test when fhevm-go will be ready.

lib/Impl.sol
```diff
@@ -63,6 +63,8 @@ interface FhevmLib {
 
     function fheIfThenElse(uint256 control, uint256 ifTrue, uint256 ifFalse) external pure returns (uint256 result);
 
+    function fheArrayEq(uint256[] memory larray, uint256[] memory rarray) external pure returns (uint256 result);
```
dartdart26: Nit: May rename to `lhs` and `rhs` as we already use it for other ops.

lib/TFHE.sol
```diff
@@ -5262,6 +5262,71 @@ library TFHE {
         return euint64.wrap(Impl.select(ebool.unwrap(control), euint64.unwrap(a), euint64.unwrap(b)));
     }
 
+    function eq(euint4[] memory a, euint4[] memory b) internal pure returns (ebool) {
+        require(a.length != b.length, "Both arrays are not of the same size.");
```
dartdart26: As discussed offline, needs to be `==` in all overloads. :)

dartdart26 (approved):
Looks good!

The only thing that we need to fix is the requires as mentioned in one of my previous messages.

codegen/templates.ts
```diff
@@ -534,6 +536,23 @@ function tfheSelect(inputBits: number): string {
...
+  return `
+    function eq(euint${inputBits}[] memory a, euint${inputBits}[] memory b) internal pure returns (ebool) {
+        require(a.length == b.length, "Both arrays are not of the same size.");
+        uint256[] memory lhs = new uint256[](a.length);
+        uint256[] memory rhs = new uint256[](b.length);
+        for (uint i = 0; i < a.length; i++) {
+          lhs[i] = euint${inputBits}.unwrap(a[i]);
+        }
+        for (uint i = 0; i < b.length; i++) {
+          rhs[i] = euint${inputBits}.unwrap(a[i]);
```
dartdart26:
```suggestion
          rhs[i] = euint${inputBits}.unwrap(b[i]);
```

test/tfheOperations/manual.ts
```diff
@@ -28,6 +28,41 @@ describe('TFHE manual operations', function () {
     this.instances = instances;
   });
 
+  it('array of value eq euint4,euint4 true', async function () {
```
jatZama: Can you add tests that would also return false?

examples/tests/TFHEManualTestSuite.sol
```diff
@@ -5,6 +5,71 @@ import "../../abstracts/Reencrypt.sol";
 import "../../lib/TFHE.sol";
 
 contract TFHEManualTestSuite is Reencrypt {
+    function test_eq_array_4(bytes calldata a, bytes calldata b) public view returns (bool) {
```
jatZama: Ideally I would also add here tests that take arrays with different elements, to avoid bad surprises.

immortal-tofu: wdym?

jatZama: You are currently only testing arrays with duplicated elements.

jatZama (approved):

PR #391, opened 2024-06-10
docs: update structure

immortal-tofu:


docs/guides/decrypt.md
```diff
@@ -0,0 +1,184 @@
...
+
+function getParamsEUint32(uint256 requestID) internal;
+
+function getParamsEUint64(uint256 requestID) internal;
+
+function getParamsEAddress(uint256 requestID) internal;
+
+function getParamsAddress(uint256 requestID) internal;
+
+function getParamsUint(uint256 requestID) internal;
```
jatZama: This function has been renamed `getParamsUint256`

docs/guides/decrypt.md
```diff
@@ -0,0 +1,184 @@
...
+
+function addParamsEUint32(uint256 requestID, euint32 _euint32) internal;
+
+function addParamsEUint64(uint256 requestID, euint64 _euint64) internal;
+
+function addParamsEAddress(uint256 requestID, address _eaddress) internal;
+
+function addParamsAddress(uint256 requestID, address _address) internal;
+
+function addParamsUint(uint256 requestID, uint256 _uint) internal;
```
jatZama: This function has been renamed `addParamsUint256`

docs/guides/decrypt.md
```diff
@@ -0,0 +1,184 @@
...
+
+  constructor() {
+      xUint32 = TFHE.asEuint32(32);
+  }
+
+  function requestUint32(uint32 input1, uint32 input2) public {
+      euint32[] memory cts = new euint32[](1);
+      cts[0] = xUint32;
+      uint256 requestID = Gateway.requestDecryption(cts, this.callbackUint32.selector, 0, block.timestamp + 100);
+      addParamsUint(requestID, input1);
```
jatZama: Same, rename it addParamsUint256

docs/guides/decrypt.md
```diff
@@ -0,0 +1,184 @@
...
+  uint32 public yUint32;
+
+  constructor() {
+      xUint32 = TFHE.asEuint32(32);
+  }
+
+  function requestUint32(uint32 input1, uint32 input2) public {
+      euint32[] memory cts = new euint32[](1);
+      cts[0] = xUint32;
+      uint256 requestID = Gateway.requestDecryption(cts, this.callbackUint32.selector, 0, block.timestamp + 100);
```
jatZama: You are missing the last boolean argument, which should be false in this case.
Please, see updated example with latest function signature here : https://github.com/zama-ai/coprocessor-contracts/blob/main/contracts/examples/TestAsyncDecrypt.sol

docs/guides/decrypt.md
```diff
@@ -0,0 +1,184 @@
...
+
+import "../lib/TFHE.sol";
+import "../gateway/GatewayCaller.sol";
+
+contract TestAsyncDecrypt is GatewayCaller {
+  euint32 xUint32;
+  uint32 public yUint32;
+
+  constructor() {
+      xUint32 = TFHE.asEuint32(32);
```
jatZama, line 123: You forgot to allow xUint32 in the constructor.
Please, see updated example here : https://github.com/zama-ai/coprocessor-contracts/blob/main/contracts/examples/TestAsyncDecrypt.sol

jatZama (changes requested):

docs/guides/decrypt.md
immortal-tofu: I didn't change a line from decrypt documentation. I'll take a look

jat9292: yes but as I said, you should, since API changed. Also by the way, cts should now be an array of uint256, not euint32[]. See my example file.

docs/fundamentals/acl.md
```diff
@@ -0,0 +1,87 @@
...
+  random = TFHE.randEuint64();
+
+  // Permanently store the temporary access for this ciphertext.
+  TFHE.allow(random, address(this));
+}
+```
+
+## Security best practice: isSenderAllowed()
+
+When a function receives a ciphertext, it needs to verify that the sender also has access to this ciphertext. This is important because otherwise, a contract could send any ciphertext authorized for the contract and potentially exploit the function to retrieve the value.
```
jat9292: Maybe I would be more specific here: "When a function receives a ciphertext handle type such as ebool, euintXX, ebytes256 or eaddress" to not confuse with the ehandle type for user inputs.

docs/guides/decrypt.md
```diff
@@ -0,0 +1,188 @@
...
+  bool public yBool;
+
+  constructor() {
+      xBool = TFHE.asEbool(true);
+      TFHE.allow(xBool, address(this));
+  }
+
+  function requestBool() public {
+    ebool[] memory cts = new ebool[](1);
+    cts[0] = xBool;
```
jatZama: Here, again, as I said previosuly , you need to pass a uint256[] array, so please use ```        uint256[] memory cts = new uint256[](1);
        cts[0] = Gateway.toUint256(xBool);```
See https://github.com/zama-ai/coprocessor-contracts/blob/main/contracts/examples/TestAsyncDecrypt.sol#L57C1-L58C43

docs/guides/decrypt.md
```diff
@@ -0,0 +1,188 @@
...
+    return yBool;
+  }
+```
+
+Note that an [`GatewayContract`](../../gateway/GatewayContract.sol) contract is already predeployed on the fhEVM testnet, and a default relayer account is added through the specification of the environment variable `PRIVATE_KEY_GATEWAY_RELAYER` in the `.env` file. Relayers are the only accounts authorized to fulfil the decryption requests. However `GatewayContract` would still check the KMS signature during the fulfilment, so we trust the relayer only to forward the request on time, a rogue relayer could not cheat by sending fake decryption results (the KMS signature is in the works).
+
+The interface of the `Gateway.requestDecryption` function from previous snippet is the following:
+
+```solidity
+function requestDecryption(
```
jatZama, line 38: The correct signature is now : ```    function requestDecryption(
        uint256[] calldata ctsHandles,
        bytes4 callbackSelector,
        uint256 msgValue,
        uint256 maxTimestamp,
        bool passSignaturesToCaller
    )```

docs/guides/decrypt.md
```diff
@@ -0,0 +1,188 @@
...
+  uint32 public yUint32;
+
+  constructor() {
+      xUint32 = TFHE.asEuint32(32);
+      TFHE.allow(xUint32, address(this));
+  }
+
+  function requestUint32(uint32 input1, uint32 input2) public {
+      euint32[] memory cts = new euint32[](1);
+      cts[0] = xUint32;
```
jatZama: Here also, use  Gateway.toUint256 please

jatZama (approved):

PR #403, opened 2024-06-18
feat: add support for list-based inputs

dartdart26:
More info: https://github.com/zama-ai/fhevm-go/pull/120

API changes will be described in the documentation.

More tests and examples are to be done in a future commit.


examples/CMUX.sol
```diff
@@ -12,10 +12,10 @@ contract SELECT is Reencrypt {
     constructor() {}
 
     // Set result = if control { ifTrue } else { ifFalse }
-    function select(bytes calldata controlBytes, bytes calldata ifTrueBytes, bytes calldata ifFalseBytes) public {
-        ebool control = TFHE.asEbool(controlBytes);
-        euint8 ifTrue = TFHE.asEuint8(ifTrueBytes);
-        euint8 ifFalse = TFHE.asEuint8(ifFalseBytes);
+    function select(einput controlBytes, einput ifTrueBytes, einput ifFalseBytes, bytes calldata input) public {
```
jatZama: instead of second argument being called input, I would prefer `inputProof` as this will be clearer later when we will append the proof to it.

examples/EncryptedERC20.sol
```diff
@@ -51,8 +51,8 @@ contract EncryptedERC20 is Ownable2Step {
     }
 
     // Transfers an encrypted amount from the message sender address to the `to` address.
-    function transfer(address to, bytes calldata encryptedAmount) public virtual {
-        transfer(to, TFHE.asEuint64(encryptedAmount));
+    function transfer(address to, einput amount, bytes calldata input) public virtual {
```
jatZama: same here, in corpocessor i called it `inputProof`as well

examples/CMUX.sol
dartdart26: Will fix it.

examples/EncryptedERC20.sol
dartdart26: Yeah, will fix.

immortal-tofu (approved):

examples/CMUX.sol
dartdart26: Fixed everywhere.

PR #405, opened 2024-06-19
docs: add fhevmjs references

immortal-tofu:


docs/guides/reencryption.md
```diff
@@ -34,7 +34,7 @@ const USER_ADDRESS = accounts[0];
 // Create a fhevmjs instance using the Zama's network and the Zama's Gateway
 const instance = await createInstance({
```
jatZama, line 35: You forgot to include chainId argument which is mandatory here.

docs/fundamentals/types/conditions.md
```diff
@@ -48,6 +49,11 @@ function _transfer(address from, address to, euint32 amount) internal {
 
   // Add to the balance of `to` and subract from the balance of `from`.
   balances[to] = balances[to] + TFHE.select(canTransfer, amount, TFHE.asEuint32(0));
```
jatZama: + operator is not overloaded anymore, so use TFHE.add instead.
And same for substraction 3 lines after to be replace by TFHE.sub

docs/getting_started/ethereum.md
```diff
@@ -5,5 +5,6 @@ You can deploy any fhEVM contract on Ethereum Sepolia. In this scenario, our Cop
 Ensure you import the appropriate configuration before deploying to Ethereum:
 
 ```solidity
-import `fhevm/networks/ethereum.sol`;
+import "fhevm/networks/ethereum.sol";
```
jatZama: Can you clarify what ethereum.sol should contain?
Fyi Sepolia has different chainID than ethereum, which also have different chainID than local geth coprocessor node, and coprocessor contracts (FHEVMCOprocessor, ACL, KMSVerifier, Gateway, etc) address might vary between each chain as well, not usre how will this be handled in a clean way.

immortal-tofu: The idea is to import a config file in the network directory. This config file would include address of the coprocessor contract.
I can rename it `sepolia.sol` 👍

jatZama: I am sorry but this is not how it works at the moment, and I don't see a clear way to do your implementation of a unique file.
Currently when testing on a local network, i.e when doing `npm run coprocessor:start` it generates automatically several files with appropriate addresses were different contracts were deployed : ACLAddress.sol, FHEVMCoprocessorAddress.sol , etc. Like what was done for Gateway/OraclePredeploy.sol in previous fhevm0.4.0
I think it is easier to do it that way, and the day you want to deploy on mainnet, if the contracts are already predeployed, you change those files.

jatZama: i.e no need for importing an additional ethereum.sol with current repo.

immortal-tofu: Anyway, we'll make this page as "coming soon". We are not ready to ship the coprocessor publicly.
Regarding, the implementation, we discussed it with @dartdart26 and he's working on an unified TFHE.sol/Impl.sol library in the coming weeks. It won't be ready for the initial release (= the initial release will only work on L1), but we'll push a patch version to enable coprocessor usage.

jatZama: TFHE and Impl libraries are already the same for both systems, coprocessor and L1.

jatZama (approved):

PR #406, opened 2024-06-20
feat: update with latest Solidity implementation

dartdart26:
There are a number of tests that are still not rewritten with the new APIs. The EncryptedERC20 and Rand examples do work.


codegen/main.ts
```diff
@@ -18,11 +18,11 @@ function generateAllFiles() {
   writeFileSync('mocks/Impl.sol', t.implSolMock(context, operators));
   const [tfheSolSourceMock, _] = t.tfheSol(context, operators, SUPPORTED_BITS, true);
   writeFileSync('mocks/TFHE.sol', tfheSolSourceMock);
-  mkdirSync('examples/tests', { recursive: true });
-  ovShards.forEach((os) => {
-    writeFileSync(`examples/tests/TFHETestSuite${os.shardNumber}.sol`, testgen.generateSmartContract(os));
-  });
-  writeFileSync('test/tfheOperations/tfheOperations.ts', testgen.generateTestCode(ovShards));
+  // mkdirSync('examples/tests', { recursive: true });
```
immortal-tofu, line 21: Commented on purpose?

mocks/Impl.sol
```diff
@@ -119,7 +119,7 @@ library Impl {
         return y;
     }
 
-    function cmux(uint256 control, uint256 ifTrue, uint256 ifFalse) internal pure returns (uint256 result) {
+    function select(uint256 control, uint256 ifTrue, uint256 ifFalse) internal pure returns (uint256 result) {
```
immortal-tofu, line 122: We already have select. This cmux was still there for migration in 0.4.0, I think you can delete it now.

jatZama: I think we should remove everything related to mock anyways. Mock used to have quite different behavior than real fhevm with previous system, for instance there was no possible detection of an unitialized handle and no tracking of the implicit ACL. 
On coprocessor-contracts repo I developped a much more closer mock version to the real system now (and still test-deving it atm) this would be far better system, and same mocked mode could be used on fhevm. wdyt @immortal-tofu

immortal-tofu: Mocks + hardhat will be the default way to test your contracts. Any optimizations is more than welcome :)
This said, can we move your mocks in fhevm project? Otherwise I think it's better to keep it until we have your update version here.

jatZama: The problem is that I am developping using Geth version of coprocessor since I could run it in a docker image. I tried to use the fhevm node but I was unsuccessful last week because of a lack of docker image and build is complex. When docker image will be available for fhevm as well I will simply retry it here.

codegen/main.ts
dartdart26: Yes, tests haven't been rewritten without decrypt yet.

jatZama (approved):
I checked thoroughly equivalence with Coprocessor contracts, and we are 100% identical, except from nit-picking detail on comments (consistency between encrypted boolean / eaddress / euintXX when doing trivial encrypt of different types) and different import relative path in Gateway library and contracts.
As said on Slack, we still need to decide on how to properly setup the hardcoded constants inside Gateway library, but this could be done in another PR. I am approving this for now.

immortal-tofu (approved):

PR #407, opened 2024-06-21
feat: pass a contract address to verifyCiphertext

dartdart26:
Will be needed in the future to verify the input ZKPoK.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,18 +116,15 @@ import "fhevm/lib/TFHE.sol";
 contract Counter {
   euint32 counter;
 
-  function add(bytes calldata encryptedValue) public {
-    euint32 value = TFHE.asEuint32(encryptedValue);
-    counter = counter + value;
-  }
-
-  function getCounter(bytes32 publicKey) returns (bytes memory) {
-    return TFHE.reencrypt(counter, publicKey);
+  function add(einput valueInput, bytes calldata inputProof) public {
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -116,18 +116,14 @@ import "fhevm/lib/TFHE.sol";
...
-  function add(bytes calldata encryptedValue) public {
-    euint32 value = TFHE.asEuint32(encryptedValue);
-    counter = counter + value;
-  }
-
-  function getCounter(bytes32 publicKey) returns (bytes memory) {
-    return TFHE.reencrypt(counter, publicKey);
+  function add(einput valueInput, bytes calldata inputProof) public {
+    euint32 value = TFHE.asEuint32(valueInput, inputProof);
+    counter = TFHE.add(counter, value);
```
jatZama, line 121: Please add TFHE.allow(counter, address(this)); here!

jatZama (approved):
I just added a minor comment concerning the README example, otherwise all good.

README.md
dartdart26: Can you explain the reasoning?

jatZama: because counter is a state variable, so if you want to persist it in ACL after the transaction it is mandatory.

dartdart26: Discussed offline - obviously needed if further operations are to be done on the counter. Adding the allow() call.

jatZama (approved):

PR #429, opened 2024-07-04
Update hardhat.config.ts

Budalebah:
Updated import statements in hardhat.config.ts for better organization and readability. for faster migration from local files to external libraries, external package imports were introduced before local imports


Budalebah:
Check please

PR #431, opened 2024-07-05
fix: KMSVerifier, Solidity ver and silence transient warning

jatZama:
Fixes the threshold in KMSVerifier to be more gas efficient, updates Solidity version to avoid incompatibility with hardhat, and silences the transient storage warning when compiling.
Would be good to make a new release with this, as it is also impacting the documentation by changing the solidity version in the code snippets examples fyi @immortal-tofu


leventdem (approved):
+1

immortal-tofu:
You need to merge it on `release/0.5.x` not on main branch.
Also, can you remove `abstracts` folder and remove `!abstracts/**` from `.npmignore`? It would be a good occasion before the release.
Final comment: when you update version you can do it by:
- changing the value in package.json and do a npm install: the npm install will also update package-lock.json
- do `npm version 0.5.2` which do both

PR #452, opened 2024-07-30
chore: fix small typo in the documentation about re-encryption

fd0r:
Following a discussion at the office with @leventdem we noticed a small typo in the documentation. Indeed even in the case of re-encryption the gateway receives the handle and not the cipher-text.

c.f. https://github.com/zama-ai/fhevm/blob/914699f9695c6f5205ec011be1155189f8abba5b/test/encryptedERC20/EncryptedERC20.ts#L76-L91

### DIFF
--- a/.github/workflows/linkchecker.yml
+++ b/.github/workflows/linkchecker.yml
@@ -18,3 +18,4 @@ jobs:
         with:
           use-quiet-mode: "yes"
           use-verbose-mode: "yes"
+          config-file: "mlc_config.json"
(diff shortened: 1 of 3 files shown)


immortal-tofu:
Can you cherry-pick your commit also on `release/0.5.x` ? Otherwise LGTM

fd0r:
> Can you cherry-pick your commit also on release/0.5.x ? Otherwise LGTM

Yes will do!

What should we do about the link checker?

fd0r:
I had to break down the url in the commit message because of the line-length limit in commitlint.

(https://github.com/conventional-changelog/commitlint/issues/2112)

immortal-tofu (approved):

.github/workflows/linkchecker.yml
```diff
@@ -18,3 +18,4 @@ jobs:
         with:
           use-quiet-mode: "yes"
           use-verbose-mode: "yes"
+          config-file: "mlc_config.json"
```
immortal-tofu, line 21: Is it normal?

fd0r: Yes this is to avoid ci crashes when the fhevm documentation raises a 429.

I can remove the last commit if you prefer

immortal-tofu:
@fd0r you can merge if it's still relevant

PR #453, opened 2024-07-31
chore: fix small typo in the documentation about re-encryption

fd0r:
c.f. https://github.com/zama-ai/fhevm/blob/914699f9695c6f5205ec011be1155189f8abba5b/test/encryptedERC20/EncryptedERC20.ts#L76-L91

Following a discussion at the office with @leventdem we noticed a small typo in the documentation. Indeed even in the case of re-encryption the gateway receives the handle and not the cipher-text.


fd0r:
I had to break down the url in the commit message because of the line-length limit in commitlint.

(https://github.com/conventional-changelog/commitlint/issues/2112)

immortal-tofu (approved):

PR #457, opened 2024-08-06
feat: new mocked mode

jatZama:


test/instance.ts
```diff
@@ -1,28 +1,46 @@
-import fhevmjs, { clientKeyDecryptor, getCiphertextCallParams, getPublicKeyCallParams } from 'fhevmjs';
+import fhevmjs, { clientKeyDecryptor, getCiphertextCallParams } from 'fhevmjs';
```
immortal-tofu: Import `createInstance` instead of the `default` with `fhevmjs`.

immortal-tofu:
LGTM

immortal-tofu (approved):

test/instance.ts
jatZama: this would not work because I am defining a createInstance function (same name) in this file, which is also needed in current tests. Otherwise I would need to find a new name and change all the tests, what should I do?

immortal-tofu: Simply use
`import { createInstance as createFhevmInstance, clientKeyDecryptor, getCiphertextCallParams } from 'fhevmjs';`

PR #479, opened 2024-08-27
feat: FHEPayment v0

jatZama:


lib/Impl.sol
```diff
@@ -275,4 +282,16 @@ library Impl {
     function isAllowed(uint256 handle, address account) internal view returns (bool) {
         return IACL(aclAdd).isAllowed(handle, account);
     }
+
+    function depositForAccount(address account, uint256 amount) internal {
```
immortal-tofu: I'd remove that. This function won't be used for any stack. Payment will be handled separately, not in TFHE library.

jatZama: My last commit addressed this concern, since I moved all FHEPayment-related functions such as deposit, withdraw and getBalance etc from TFHE.sol to a Payment.sol library.

immortal-tofu (approved):
LGTM!

PR #513, opened 2024-09-19
fix: Use #!/usr/bin/env bash rather than #!/bin/bash

silasdavis:
As an act of selfless kindness to non FHS-compliant OSes (like NixOS)

This should be strictly more portable than the current shebang


immortal-tofu:
Can you fix your commit? Other LGTM :)

silasdavis:
I think it was `s/Use/use` :), if so, done

immortal-tofu:
Scripts changed, if you can resolve conflicts. Sorry for that.

PR #535, opened 2024-09-27
feat: set null FHE gas prices +  remove Payment lib

jatZama:


lib/FHEPayment.sol
```diff
@@ -29,9 +29,9 @@ contract FHEPayment is UUPSUpgradeable, Ownable2StepUpgradeable {
     uint256 private constant PATCH_VERSION = 0;
     address public constant tfheExecutorAddress = tfheExecutorAdd;
 
-    uint256 private constant FHE_GAS_BLOCKLIMIT = 10_000_000;
-    uint256 private constant MIN_FHE_GASPRICE = 10_000_000; // minimum of 0.01 Gwei
-    uint256 private constant FHE_GASPRICE_NATIVE_RATIO = 1000; // fhe gas price is set to 0.1% of native gas price (if above minimum)
+    uint256 private constant FHE_GAS_BLOCKLIMIT = 10_000_000; //
```
immortal-tofu: Useless `//`

lib/FhevmLib.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
```
immortal-tofu: What is this file exactly?

immortal-tofu:
You want to keep https://github.com/zama-ai/fhevm/blob/main/payment/Payment.sol or not?

lib/FhevmLib.sol
jatZama: this was an artefact from old codegen, when we needed the precompile, i removed it.

lib/FHEPayment.sol
jatZama: artefact from codegen, removed it as well.

jatZama:
> You want to keep https://github.com/zama-ai/fhevm/blob/main/payment/Payment.sol or not?

Yes of course, this is still a lot of work done here, and tests for this library as still here and well covered, if we want to reactivate non-free FHE operations one day. Imo this should be kept for the moment, but we are not obliged to publish it, we could add it in the `.npmignore` since it is in its own folder. With actual content of `.npmignore` it won't be published anyways.

immortal-tofu:
No problem to keep it, it was regarding the title 👍

immortal-tofu (approved):

PR #550, opened 2024-10-04
Adding Husky and Commitlint

poppyseedDev:
I noticed my pull requests a lot of times weren't passing the following checks:
 - pull request tests/build -> fails to run `npm run prettier:check`
 - Run `commitlint` on PR 
 
 Therefore I decided to add husky to automatically enforce these rules on every commit.


immortal-tofu (approved):

commitlint.config.js
PacificYield, line 1: It may be best to use it as a TS file (so it would need to be renamed to `.commitlintrc.ts`). You would need to install the types (@commitlint/types).

```
import type { UserConfig } from "@commitlint/types";

const Configuration: UserConfig = {
  extends: ["@commitlint/config-conventional"],
  rules: {
    "type-enum": [2, "always", ["ci", "chore", "docs", "ticket", "feat", "fix", "perf", "refactor", "revert", "style"]],
  },
};

export default Configuration;
```

.husky/commit-msg
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
```
PacificYield: These can be removed

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
```
PacificYield: Same here

.husky/commit-msg
poppyseedDev:
```suggestion
```

.husky/pre-commit
poppyseedDev:
```suggestion
```

commitlint.config.js
PacificYield: It needs to also include `"test"`

PR #571, opened 2024-10-10
fix: fix relayer typo

dd23:


jatZama:
this will be part of #572

PR #576, opened 2024-10-16
Minor fixes when running in localCoprocessor mode

manoranjith:
- use gateway addr from parsedEnv
- load relayer priv key only in mock


jatZama:
Please run `npm run prettier` since CI is failing otherwise

jatZama (approved):

PR #585, opened 2024-10-21
Unauth reencrypt test

jatZama:


PacificYield (approved):
LGTM, it may be worth adding a comment about the reasons why hardhat chai matchers cannot be used

PR #587, opened 2024-10-23
feat: TFHEExecutor events variant with faster mock

jatZama:


lib/TFHEExecutor.events.sol
```diff
@@ -0,0 +1,679 @@
...
+    function fheDiv(uint256 lhs, uint256 rhs, bytes1 scalarByte) external virtual returns (uint256 result) {
+        require(scalarByte & 0x01 == 0x01, "Only fheDiv by a scalar is supported");
+        require(rhs != 0, "Could not divide by 0");
+        uint256 supportedTypes = (1 << 1) + (1 << 2) + (1 << 3) + (1 << 4) + (1 << 5) + (1 << 6) + (1 << 8);
+        requireType(lhs, supportedTypes);
+        uint8 lhsType = typeOf(lhs);
+        bytes1 scalar = scalarByte & 0x01;
+        fhePayment.payForFheDiv(msg.sender, lhsType, scalar);
+        result = binaryOp(Operators.fheDiv, lhs, rhs, scalar, lhsType);
+        emit FheDiv(lhs, rhs, scalarByte, result);
```
PacificYield, line 263: It may be worth to consider inheriting from `TFHEExecutor.sol` to avoid duplicating the codebase since only the events are new

PacificYield (approved):

PR #590, opened 2024-10-25
feat: refactor importing fhevm-core-contracts

jatZama:


melanciani (approved):

immortal-tofu (commented):
I did a quick review. I think most of the task should be moved to fhevm-backend because writing stuff in node_modules is not a good practice, especially because on `npm install` any change to `node_modules` will be removed.
Not sure if it's problematic to move these tasks (upgradeProxy, etherscanveriufy, taskTFHE, taskDeploy, ...)

PR #596, opened 2024-10-29
fix: update kms signer0 address

tawadaa:


leventdem (commented):
It seems you docker image was old even with the right tag, so this PR is not relevant, we can remove it

tawadaa:
`kms-service-dev` image changes depending on build platform, this PR is closed until we dynamically set the signer in L1 demo.

PR #608, opened 2024-11-15
feat: add default configs for Gateway and FHEVM

PacificYield:
This PR includes default config files (in Solidity) that can be inherited to automatically set the addresses to use (1) Zama's FHEVM, (2) Zama's Gateway.

In the future, it may be better to create a standardized repo for any service provider to add their config information.

These configs are meant to be imported such as:

```solidity
import {MockZamaFHEVMConfig} from "fhevm/config/ZamaFHEVMConfig.sol";
```

and can be inherited such as:

```solidity
contract MyEncryptedERC20 is MockZamaFHEVMConfig, EncryptedERC20 {
    /// My code
}

### DIFF
--- a/config/ZamaFHEVMConfig.sol
+++ b/config/ZamaFHEVMConfig.sol
@@ -1,0 +1,70 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+pragma solidity ^0.8.24;
+
+import {FHEVMConfig, TFHE} from "../lib/TFHE.sol";
+
+/**
+ * @title   ZamaFHEVMConfig.
+ * @notice  This library returns the TFHE config for different networks
+ *          with the contract addresses for
+ *          (1) ACL, (2) TFHEExecutor, (3) FHEPayment, (4) KMSVerifier,
+ *          which are deployed & maintained by Zama.
+ */
+library ZamaFHEVMConfig {
+    function getMockConfig() internal pure returns (FHEVMConfig.FHEVMConfigStruct memory) {
+        return
+            FHEVMConfig.FHEVMConfigStruct({
+                ACLAddress: 0x339EcE85B9E11a3A3AA557582784a15d7F82AAf2,
+                TFHEExecutorAddress: 0x596E6682c72946AF006B27C131793F2b62527A4b,
+                FHEPaymentAddress: 0x6d5A11aC509C707c00bc3A0a113ACcC26c532547,
+                KMSVerifierAddress: 0x208De73316E44722e16f6dDFF40881A3e4F86104
+            });
+    }
+
+    function getSepoliaConfig() internal pure returns (FHEVMConfig.FHEVMConfigStruct memory) {
+        /// TODO
+    }
+
+    function getEthereumConfig() internal pure returns (FHEVMConfig.FHEVMConfigStruct memory) {
+        /// TODO
+    }
+}
+
+/**
+ * @title   MockZamaFHEVMConfig.
+ * @dev     This contract can be inherited by a contract wishing to use these contracts on the mock
+ *          environment provided by Zama.
+ *          Other providers may offer similar contracts deployed at different addresses.
+ *          If you wish to use them, you should rely on the instructions from these providers.
+ */
+contract MockZamaFHEVMConfig {
+    constructor() {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: config/DefaultFHEVMConfig.sol [hunk 1]
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+pragma solidity ^0.8.24;
+
+import "../lib/TFHE.sol";
```
immortal-tofu: Default config should not be in TFHE.sol. Keep it here directly.

config/DefaultGatewayConfig.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+pragma solidity ^0.8.24;
+
+import "../gateway/lib/Gateway.sol";
```
immortal-tofu: Same.

config/DefaultFHEVMConfig.sol
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * @dev Address of the FHEPayment contract from Zama for mock purposes.
+ */
+address constant FHE_PAYMENT_MOCK = 0x6d5A11aC509C707c00bc3A0a113ACcC26c532547;
+
+/**
+ * @dev Address of the KMSVerifier contract from Zama for mock purposes.
+ */
+address constant KMS_VERIFIER_MOCK = 0x208De73316E44722e16f6dDFF40881A3e4F86104;
```
immortal-tofu: If we embed config for Sepolia, Mainnet, Polygon, ... maybe it's not relevant to export all const? Having access to ZamaFHEVMConfig is enough?

config/DefaultFHEVMConfig.sol
```diff
@@ -0,0 +1,90 @@
+// SPDX-License-Identifier: BSD-3-Clause-Clear
```
immortal-tofu: Rename `DefaultFHEVMConfig` to `ZamaFHEVMConfig` since it's not only default.

config/DefaultFHEVMConfig.sol
```diff
@@ -0,0 +1,90 @@
...
+}
+
+/**
+ * @title   SepoliaZamaFHEVMConfig.
+ * @dev     This contract can be inherited by a contract wishing to use the FHEVM contracts provided by Zama
+ *          on the Sepolia network.
+ *          Other providers may offer similar contracts deployed at different addresses.
+ *          If you wish to use them, you should rely on the instructions provided by these providers.
+ */
+contract MainnetZamaFHEVMConfig {
```
immortal-tofu: Mainnet is not very relevant if we run a Coprocessor on different chains (Ethereum, Polygon, ...). Maybe we should use `ETH` if it's ETH mainnet?

config/DefaultGatewayConfig.sol
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * @title   MockZamaGatewayConfig.
+ * @notice  This config contains the mock Gateway config with the address for the Gateway
+ *          contract deployed & maintained by Zama.
+ * @dev     This contract can be inherited by a contract wishing to use the Gateway service
+ *          provided by Zama.
+ *          Other providers may offer other Gateways deployed at different addresses.
+ *          If you wish to use them, you should rely on the instructions provided by these providers.
+ */
+contract MockZamaGatewayConfig {
```
immortal-tofu: Maybe we should use same structure (library containing addresses for different networks?)

config/DefaultFHEVMConfig.sol
PacificYield: Yes, I think so

config/DefaultGatewayConfig.sol
PacificYield: Adjusted!

config/DefaultFHEVMConfig.sol
PacificYield: Yes, I changed to `EthereumZamaFHEVMConfig`

config/DefaultFHEVMConfig.sol
PacificYield: Good catch

immortal-tofu (approved):

PR #624, opened 2024-11-28
feat: added scripts for reading and removing signers

jatZama:


fd0r (approved):

immortal-tofu (changes requested):
This PR should be on fhevm-backend

immortal-tofu (approved):

PR #635, opened 2024-12-03
docs: add new whitepaper

immortal-tofu:


jatZama:
Is it normal to not have any author names? just wondering if this was not missed since original version had names on it, and it might be needed for an official publication.

jatZama (approved):

PR #658, opened 2024-12-16
refactor: add custom errors for TFHE

PacificYield:
This PR replaces require statements for custom errors in the `TFHE.sol` library.

### DIFF
--- a/codegen/templates.ts
+++ b/codegen/templates.ts
@@ -265,6 +265,16 @@ ${commonSolLib()}
  *          that interact with TFHE.
  */
 library TFHE {
+
+/// @notice Returned if the input's length is greater than 64 bytes.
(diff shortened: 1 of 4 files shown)


codegen/templates.ts
```diff
@@ -265,6 +265,16 @@ ${commonSolLib()}
  *          that interact with TFHE.
  */
 library TFHE {
+
+/// @notice Returned if the input's length is strictly greater than 64 bytes.
```
jatZama: you dont need to put "strictly" in English fyi

jatZama (approved):

codegen/templates.ts
PacificYield: Adjusted!

PR #668, opened 2024-12-20
End of year 2024 break

yuxizama:
Indeed to show the block on the frontend, we should merge it into 0.6 branch. 
(It's ok to merge doc changes into release branch, that's what we do each time for release also) @bcm-at-zama

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -17,6 +17,9 @@ layout:
 
 # Welcome to fhEVM
 
+{% hint style="warning" %} Please note that our testnet support will be limited during the holiday season, from 23th December to 5th January. During this time, the HackathonNet Sepolia network is not be actively maintained.
+{% endhint %}
+
 ## Get started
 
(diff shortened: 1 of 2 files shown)


bcm-at-zama (approved):
approved

bcm-at-zama:
not sure what's the commit e69cbd8c5b6a74a43df9c29d890a9d04a2160deb here

immortal-tofu:
> not sure what's the commit [e69cbd8](https://github.com/zama-ai/fhevm/commit/e69cbd8c5b6a74a43df9c29d890a9d04a2160deb) here
Because you try to merge your branch coming from master to `release/0.6.x`
The best way is to start from `release/0.6.x` (and cherry pick commits if needed)

Since JAT did exactly the same mistake and merged the whole `main` branch in release/0.6.x, you only add the merge commit by doing the same mistake.

Let's merge it as it is.

PR #674, opened 2024-12-23
feat: updated to new core contracts prerelease

jatZama:


immortal-tofu (approved):

lib/Impl.sol
```diff
@@ -70,6 +70,10 @@ interface IACL {
     function allowForDecryption(uint256[] memory handlesList) external;
 }
 
+interface IInputVerifier {
```
PacificYield, line 77:
```
/**
 * @title IInputVerifier
 * @notice  This interface contains the only function required from InputVerifier.
 */
```

Maybe something like this

PacificYield (approved):

PR #698, opened 2025-01-22
feat: updated to new fhevm-core-contracts prerelease

jatZama:


package.json
```diff
@@ -72,7 +70,7 @@
     "eslint": "^8.28.0",
     "eslint-config-prettier": "^8.5.0",
     "ethers": "^6.8.0",
-    "fhevm-core-contracts": "0.7.0-3",
+    "fhevm-core-contracts": "^0.7.0-4",
```
immortal-tofu, line 73: Is it needed to bump? You don't modify the contracts. We should avoid (pre)publish for no reason, so I'd suggest removing the bump from this PR.

jatZama: yes it is absolutely needed, since I modified the events in TFHEExecutor contracts I had to update the mock here (on fhevm) also

immortal-tofu (approved):

PR #717, opened 2025-03-06
refactor: bytes32 handles

PacificYield:
This PR updates the codegen (and its lib outputs) to support bytes32 types for handles/ciphertexts to match the updated codegen from `httpz-backend`.


lib/TFHE.sol
```diff
@@ -1467,7 +1467,7 @@ library TFHE {
         if (!isInitialized(b)) {
             b = asEuint4(0);
         }
-        return euint4.wrap(Impl.mul(euint4.unwrap(b), uint256(a), true));
+        return euint4.wrap(Impl.mul(euint4.unwrap(b), bytes32(uint256(a)), true));
```
immortal-tofu: a is a `uint8` and in this context, `uint256` is not a handle but a plaintext value:

eg: multiply a ciphertext by 3 `TFHE.mul(3, ct);`

I think you have this mistake in many place.

PacificYield: As discussed offline, we can keep it as it is since it would require duplicating most functions in `TFHEExecutor` otherwise.

jatZama: I did not understand your answer @PacificYield , @immortal-tofu is right, this second argument should NOT be a bytes32

jatZama: Ok I think I understood, what is the status on this? Did @immortal-tofu decide to keep the conversion to bytes32 here to avoid duplicated function for the scalar type?

codegen/utils.ts
```diff
@@ -1,4 +1,45 @@
+/**
+ * Generates a string representing a uint type with the specified number of bits.
+ *
+ * @param bits - The number of bits for the uint type. Must be greater than 8.
+ * @returns A string representing the uint type with the specified number of bits.
+ * @throws Will throw an error if the number of bits is less than or equal to 8.
+ */
 export const getUint = (bits: number) => {
-  if (bits <= 8) return 'uint8';
+  if (bits < 8) throw new Error('Bits must be greater than 8');
```
immortal-tofu, line 9: You removed the `<=`. Is it ok?

immortal-tofu (commented):
Since most of the code come from backend, I think it's fine. Still I let @jatZama take a look.

codegen/utils.ts
PacificYield: Yes, since it doesn't support euint4 anymore

jatZama:
I noticed you are still using a subdirectory called fhevmTemp containing the constant addresses (all null), which are needed for compilation, but I think this artifacts could be:
1/ first renamed to "httpzTemp" or something similar (removing fhevm reference) and update all corresponding paths in code.
2/ then removed from the repo, and generated on-the-fly in the test prehook in the hardhat config, wdyt? Maybe it is needed elsewhere, probably for the (pre)release workflows, in this case it coud also be generated automatically.

immortal-tofu:
I think the end goal is mainly to remove these artifacts and have proper unit testing of the library, so I wouldn't spend too much time on renaming things right now.

.env.example
```diff
@@ -1,13 +1,27 @@
...
+ZKPOK_MANAGER_ADDRESS="0x812b06e1CDCE800494b79fFE4f925A504a9A9810"
+NUM_KMS_NODES="1"
+KMS_THRESHOLD="1"
+NUM_COPROCESSORS="1"
+ETHERSCAN_API_KEY="XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
+RPC_URL="http://anvil-node:8545"
+HARDHAT_NETWORK="staging"
+
+
+PRIVATE_KEY_KMS_SIGNER_0="30d45b1c5a771e20d0ec15097c3b6ac7153bc1992bc78c42af37725dd93f096a" # accounts[11]
```
jatZama: Why are you keeping all private keys for kms signers and coprocessors? In `httpz-backend` with new deployment script we only need the addresses, I think you can remove them as well, see: https://github.com/zama-ai/httpz-backend/blob/main/contracts/.env.example

jatZama (approved):

jatZama:
I approved to be able to merge ASAP, but please take into consideration my comments if possible.

jatZama:
Can you please also remove this line from the .npmignore since this branch does not include the `decryption` directory anymore? https://github.com/zama-ai/httpz-solidity/blob/bytes32-handles/.npmignore#L3

jatZama (approved):

PR #745, opened 2025-04-18
feat: use latest core contracts, updated lib API, renamed to fhevm an…

jatZama:
…d fhe

This PR also fixes coverage task.

Renaming: httpz->fhevm everywhere EXCEPT the HTTPZ.sol library now called FHE.sol

Some major changes in API: 
- Replace the `einput` type by several types for unverified inputs: `externalEbool`, `externalEuint8` ... etc
- Replace `HTTPZ.asEuintXX` for verifying inputs by a unique overloaded function: FHE.fromExternal(externalEXX handle, bytes proof)`
- Add `FHE.makePubliclyDecryptable` and `FHE.isPubliclyDecryptable` functions. 
- Now all functions related to requesting an async decrypt and checking a signature in callback during decryption fulfilment have been moved to FHE.sol library. The DecryptionOracle contrat should adhere to the correct interface, i.e containing a `requestDecryption(uint256 requestID, bytes32[] calldata ctsHandles, bytes4 callbackSelector)` function. It has been discussed with @immortal-tofu to potentially add later a variant of `FHE.requestDecryption` with an additional `bytes` argument for custom calldata to be sent to a different kind of Oracle contract, but specs are not finalized yet.


package.json
```diff
@@ -68,8 +67,7 @@
     "eslint": "^8.28.0",
     "eslint-config-prettier": "^8.5.0",
     "ethers": "^6.8.0",
-    "@httpz/core-contracts": "^0.7.0-5",
-    "@httpz/sdk": "^0.7.0-13",
```
immortal-tofu, line 72: you removed @httpz/sdk and installed fhevmjs

package.json
```diff
@@ -81,15 +79,14 @@
     "prettier-plugin-solidity": "^1.1.3",
     "rimraf": "^4.1.2",
     "sha3": "^2.1.4",
-    "solidity-coverage": "0.8.12",
+    "solidity-coverage": "0.8.15",
```
immortal-tofu, line 82: Update of solidity coverage is maybe the cause of your issue :(

package.json
```diff
@@ -81,15 +79,14 @@
...
     "ts-generator": "^0.1.1",
     "ts-node": "^10.9.2",
     "typechain": "^8.2.0",
     "typescript": "^5.1.6",
-    "web3-validator": "^2.0.6"
-  },
-  "dependencies": {
+    "web3-validator": "^2.0.6",
     "@openzeppelin/contracts": "^5.0.1",
+    "@zama-fhe/oracle-solidity": "^0.0.2",
```
immortal-tofu: It should be a devDependencies

package.json
```diff
@@ -81,15 +79,14 @@
...
-    "solidity-coverage": "0.8.12",
+    "solidity-coverage": "0.8.15",
     "ts-generator": "^0.1.1",
     "ts-node": "^10.9.2",
     "typechain": "^8.2.0",
     "typescript": "^5.1.6",
-    "web3-validator": "^2.0.6"
-  },
-  "dependencies": {
+    "web3-validator": "^2.0.6",
```
immortal-tofu, line 87: Why move this package as dependency and not dev?

immortal-tofu (changes requested):

package.json
immortal-tofu: You can use `@fhevm/sdk@0.7.0-20`

package.json
```diff
@@ -1,11 +1,11 @@
 {
-  "name": "@httpz/solidity",
-  "description": "A Solidity library for interacting with HTTPZ protocol",
+  "name": "@fhevm/solidity",
+  "description": "A Solidity library for interacting with FHEVM protocol",
   "version": "0.7.0-2",
```
immortal-tofu: Maybe bump the version at the same time

test/instance.ts
```diff
@@ -1,11 +1,11 @@
+import dotenv from 'dotenv';
 import {
   clientKeyDecryptor,
   createEIP712,
   createInstance as createFhevmInstance,
   generateKeypair,
   getCiphertextCallParams,
-} from '@httpz/sdk';
-import dotenv from 'dotenv';
+} from 'fhevmjs';
```
0xalexbel: Maybe. `import type { FhevmInstance } from '@fhevm/sdk';` instead ? (see test/types.ts)

0xalexbel (approved):
Found a little typo I guess. "from 'fhevmjs'" instead  of "@fhevm/sdk" hope it helps!

package.json
jatZama: I was planning to update this import in a future PR (I said it this morning on Slack in the daily thread) because there are breaking changes in new pre-release in the sdk API, meaning I would have to update also the mocked mode, but I can do it now in the same PR if you prefer. I will also have to do same update in fhevm-backend obviously.

package.json
jatZama: Good catch, I thought I already moved it in devDependencies but since there was problem with the lock file maybe I forgot when I tried to regenerate everything related to packages. Sorry for the oversight.

package.json
jatZama: I will do this thanks

package.json
immortal-tofu: We were already using `@httpz/sdk`, so things was working, no?

test/instance.ts
jatZama: Ok but I will first have to update the mock mode and import the sdk library as explained here https://github.com/zama-ai/fhevm-solidity/pull/745#discussion_r2050546573

package.json
jatZama: I don't think this is the issue, since I am using this version of solidity-coverage in fhevm-backend repo and CI is working as expected there.

package.json
```diff
@@ -1,11 +1,11 @@
 {
-  "name": "@httpz/solidity",
-  "description": "A Solidity library for interacting with HTTPZ protocol",
+  "name": "@fhevm/solidity",
+  "description": "A Solidity library for interacting with FHEVM protocol",
```
PacificYield: I think it needs to be lowercase

PacificYield (approved):

package.json
jatZama: Not sure about this. Marketing team was still discussing wti @bcm-at-zama and @immortal-tofu correct case.

package.json
jatZama: No reencryption was not working since erc20 tests were skipped. I am 100% sure this needs to be updated if using the sdk.

package.json
jatZama: ok since Clément confirmed it was lowercase i'm going to fix it

package.json
jatZama: I just double checked, every dependency is already i devDependencies actually, maybe you misread?

package.json
jatZama: I just double checked, every dependency is already i devDependencies actually, maybe you misread?

package.json
bcm-at-zama: correct case is lowercase, yes

package.json
jatZama: fixed thanks to your PR with optional dependencies, thank you!

package.json
jatZama: done

immortal-tofu (approved):

test/instance.ts
jatZama: fixed with updated mocked userDecrypt

PR #771, opened 2025-05-12
docs: replace einput

PacificYield:


poppyseedDev (approved):

poppyseedDev:
Are these all einput examples that we have?

PacificYield:
@poppyseedDev Yes, I think so

